PR #1, opened 2019-07-20
Head content

jtrussell:
Add a head section to each page with title, description, and a couple other meta tags.

This also applies the `Layout` component in the page template instead, we can pull that out if folks would rather have `<Layout>...</Layout>` explicitly on every page.

Apologies, `pages/_app.js` has some extra noise as it had styles that got fist shakes from the auto-formatter.


alexmiddeleer:
@jtrussell I thought about maybe having layout be an opt-in because I wasn't sure if it was easy to opt-out. But I'm new to Next.js - if we want to have a weird page with its own layout is it still doable?

next/pages/card.js
```diff
@@ -17,14 +17,23 @@ export const cardQuery = gql`
...
+    {({ loading, error, data }) => {
+      if (loading) return null;
+      if (error) return <ErrorMessage message={`Error: ${error}`} />;
 
-        return <Card card={data.card} />;
-      }}
-    </Query>
-  </Layout>
+      return (
+        <>
```
alexmiddeleer: what dis?

alexmiddeleer: nvm I see https://reactjs.org/blog/2017/11/28/react-v16.2.0-fragment-support.html

jtrussell:
That change wouldn't make the layout easy to opt out of. Happy to swap that out if you prefer. I'll probably add keys to the meta tags too, and likely remove title/description from the template after thinking about it.

jtrussell:
@alexmiddeleer moved layout back to individual pages 👍.

alexmiddeleer:
Thanks 🎊

PR #2, opened 2019-07-21
Deck page

alexmiddeleer:


next/components/deck.js
```diff
@@ -0,0 +1,44 @@
...
+            <h1 className="deckName">{deck.name}</h1>
+            <CardList cards={deck.cardDecks.nodes} />
+          </>
+        );
+      }}
+    </Query>
+  );
+}
+Deck.propTypes = {
+  deck: PropTypes.object.isRequired
```
jtrussell: `PropTypes.object` is a weak guarantee - if we're expecting`deck.id`, `deck.name`, etc. to exist we might want to consider using `PropTypes.shape` instead:

```javascript
Deck.propTypes = {
  deck: PropTypes.shape({
    id: PropTypes.string,
    name: PropTypes.string,
    // etc.
  }).isRequired
}
```

PR #3, opened 2019-07-23
Add google analytics

jtrussell:
Adding GA in production.

### DIFF
--- a/next/next.config.js
+++ b/next/next.config.js
@@ -1,0 +1,6 @@
+module.exports = {
+  env: {
+    GA_TRACKING_ID: 'UA-142178156-1',
+    USE_GOOGLE_ANALYTICS: process.env.NODE_ENV === 'production'
+  }
+};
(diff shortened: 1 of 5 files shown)


next/next.config.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
+  env: {
+    GA_TRACKING_ID: 'UA-142178156-1',
```
alexmiddeleer: Is there a GA account with this tracking_id?

jtrussell: I assumed so https://trello.com/c/kTcQbXDh, perhaps this was a placeholder?

jtrussell: Not particularly sensitive but would you rather have externally managed secrets? That would be cool, probably out scope for this task.

alexmiddeleer: Yep totally

jtrussell: Right on - I'll be adding secrets as part of auth. Once that's in place we can move our GA ID there.

PR #4, opened 2019-07-24
Add dev dependencies for prettier linting with eslint

evsheffield:
I wasn't sure if folks had been doing this with global installs, but figured it would be helpful to just have things in place as dev dependencies so we can just `npm install` and go.

### DIFF
--- a/e2e/.eslintrc.json
+++ b/e2e/.eslintrc.json
@@ -2,5 +2,6 @@
   "plugins": ["cypress"],
   "env": {
     "cypress/globals": true
-  }
+  },
+  "extends": ["plugin:prettier/recommended"]
 }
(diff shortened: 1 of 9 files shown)


jtrussell:
Yay! Thanks for this 👍.

jtrussell:
This looks good to me, but perhaps @alexmiddeleer can weigh in as I'm less familiar with prettier configs.

evsheffield:
FWIW I did test it out with both Sublime and VS Code linter plugins and it seemed to be working. The only slightly funky thing is that both editors look for the node_modules to be in the project root (perhaps the plugins are configurable to change this) so rather than pull in the whole "mythgard-hub" folder I added each of the sub-projects to the workspace individually.

alexmiddeleer:
Lgtm, thanks

PR #5, opened 2019-07-26
Deck comments

alexmiddeleer:


next/components/create-comment.js
```diff
@@ -0,0 +1,106 @@
...
+    super(props);
+    this.state = { commentBody: '', deckId: props.deckId };
+
+    this.handleInputChange = handleInputChange.bind(this);
+    this.handleSubmit = this.handleSubmit.bind(this);
+    this.updateCache = this.updateCache.bind(this);
+  }
+
+  handleSubmit(e, addCommentQuery) {
+    e.preventDefault();
```
lily-pytel: Nothing major but I've taken to making sure `e` actually exists as a habit because I've had some cases where it didn't for no reason. Javascript paranoia++

next/components/comment-list.js
```diff
@@ -0,0 +1,16 @@
...
+      {comments.map(comment => (
+        <li data-cy="comment" key={comment.id}>
+          {comment.body}
+        </li>
+      ))}
+    </ul>
+  );
+}
+CommentList.propTypes = {
+  comments: PropTypes.array
```
jtrussell: YMMV but I've found I get the most value out of prop types when I verify all assumptions made in the same component. E.g.:

```
CommentList.propTypes = {
  comments: PropTypes.arrayOf(PropTypes.shape({
    id: PropTypes.string.isRequired,
    body: PropTypes.string.isRequired
  }))
}

CommendsList.defaultProps = {
  comments: []
}
```

If we broke this out and there was a separate comment component, we could move the id/body checks there and just verify we got an array here.

It can feel a little silly to verify the shape of a comment since GraphQL gives us that for free, but then GraphQL also assures us we'll get an array /shrug.

next/components/create-comment.js
```diff
@@ -0,0 +1,106 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import gql from 'graphql-tag';
+import { Mutation } from 'react-apollo';
+import ErrorMessage from './error-message';
+import { handleInputChange } from '../lib/form-utils';
+
+export const addCommentQuery = gql`
```
evsheffield: Looks like this is only being used in this file so perhaps it doesn't need to be exported atm?

next/components/create-comment.js
```diff
@@ -0,0 +1,106 @@
...
+        id
+      }
+    }
+  }
+`;
+
+class CreateComment extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = { commentBody: '', deckId: props.deckId };
```
evsheffield: Does `deckId` need to be in state or can we just use it from the props? It doesn't seem like the state value gets updated anywhere.

alexmiddeleer:
> image
> 
> I'm getting this error when clicking on a deck name. Something I need to set up?

@lily-pytel I'm not sure. I don't get that error. Do the e2e tests pass for you?

lily-pytel:
> > image
> > I'm getting this error when clicking on a deck name. Something I need to set up?
> 
> @lily-pytel I'm not sure. I don't get that error. Do the e2e tests pass for you?

I needed to run the `rebuild-db-and-restart-graphql.sh` first

PR #6, opened 2019-07-28
Feature/import deck

lily-pytel:
Trello: https://trello.com/c/XVjJdZiv/25-deck-import

- Added jest and all the configs that come with it, including .babelrc and enzyme. Not using enzyme quite yet but it's ready if we need it.
- Added page for deck builder
- Added page for import results
- Added validation for imports + tests
- No end-to-end tests 😢... I have not figured out how to make cypress work yet

There are barely any styles and I didn't make it a modal like eternal warcry because even that seemed like getting into the styles a bit too much. In short, this is very basic looks-wise

### DIFF
--- a/next/.babelrc
+++ b/next/.babelrc
@@ -1,0 +1,3 @@
+{
+  "presets": ["@babel/preset-env"]
+}
--- a/next/.babelrc
+++ b/next/.babelrc
@@ -1,3 +1,3 @@
 {
-  "presets": ["@babel/preset-env"]
+  "presets": ["next/babel"]
 }
--- a/next/components/card-list.js
+++ b/next/components/card-list.js
@@ -6,6 +6,7 @@ export default function CardList({ cards }) {
     <ul className="cardList">
       {cards.map((card, index) => (
         <li key={card.id}>
+          {card.quantity && <span>{card.quantity}</span>}
           <Link href={`/card?id=${card.id}`} key={index}>
             <a>
               {card.id}. name: {card.name}
--- a/next/components/card-list.js
+++ b/next/components/card-list.js
@@ -6,7 +6,7 @@ export default function CardList({ cards }) {
     <ul className="cardList">
       {cards.map((card, index) => (
         <li key={card.id}>
-          {card.quantity && <span>{card.quantity}</span>}
+          {card.quantity && <span>{card.quantity}</span>}&nbsp;
           <Link href={`/card?id=${card.id}`} key={index}>
             <a>
               {card.id}. name: {card.name}
--- a/next/components/imported-deck.js
+++ b/next/components/imported-deck.js
@@ -1,0 +1,33 @@
+import PropTypes from 'prop-types';
+import CardList from './card-list';
+
+export default function ImportedDeck({ importedDeck }) {
+  const { deckName, deckPath, mainDeck, sideboard } = importedDeck;
+
+  if (importedDeck.errors && importedDeck.errors.length) {
+    return (
+      <div>
+        <h2>Imported Deck</h2>
+        <div>Some errors have occurred:</div>
+        {importedDeck.errors.map((error, i) => (
+          <div key={i}>{error}</div>
+        ))}
+      </div>
+    );
+  }
+
+  return (
+    <div>
+      <h2>Imported Deck</h2>
+      <div>Name: {deckName}</div>
+      <div>Path: {deckPath}</div>
+      <h3>Main Board</h3>
+      <CardList cards={mainDeck} />
+      <h3>Sideboard</h3>
+      <CardList cards={sideboard} />
+    </div>
+  );
+}
+ImportedDeck.propTypes = {
+  importedDeck: PropTypes.object
+};
--- a/next/components/imported-deck.js
+++ b/next/components/imported-deck.js
@@ -2,7 +2,7 @@ import PropTypes from 'prop-types';
 import CardList from './card-list';
 
 export default function ImportedDeck({ importedDeck }) {
-  const { deckName, deckPath, mainDeck, sideboard } = importedDeck;
+  const { deckName, deckPath, deckPower, mainDeck, sideboard } = importedDeck;
 
   if (importedDeck.errors && importedDeck.errors.length) {
     return (
@@ -21,6 +21,7 @@ export default function ImportedDeck({ importedDeck }) {
       <h2>Imported Deck</h2>
       <div>Name: {deckName}</div>
       <div>Path: {deckPath}</div>
+      <div>Path: {deckPower}</div>
       <h3>Main Board</h3>
       <CardList cards={mainDeck} />
       <h3>Sideboard</h3>
--- a/next/components/imported-deck.js
+++ b/next/components/imported-deck.js
@@ -29,10 +29,10 @@ export default function ImportedDeck({ importedDeck }) {
 
   const sideboardElement =
     sideboard && sideboard.length ? (
-      <React.Fragment>
+      <>
         <h3>Sideboard</h3>
         <CardList cards={sideboard} />
-      </React.Fragment>
+      </>
     ) : null;
 
   return (
@@ -40,7 +40,7 @@ export default function ImportedDeck({ importedDeck }) {
       <h2>Imported Deck</h2>
       <div>Name: {deckName}</div>
       <div>Path: {deckPath}</div>
-      <div>Path: {deckPower}</div>
+      <div>Power: {deckPower}</div>
       <h3>Main Board</h3>
       <CardList cards={mainDeck} />
       {sideboardElement}
--- a/next/components/imported-deck.js
+++ b/next/components/imported-deck.js
@@ -27,6 +27,14 @@ export default function ImportedDeck({ importedDeck }) {
     );
   }
 
+  const sideboardElement =
+    sideboard && sideboard.length ? (
+      <React.Fragment>
+        <h3>Sideboard</h3>
+        <CardList cards={sideboard} />
+      </React.Fragment>
+    ) : null;
+
   return (
     <div>
       <h2>Imported Deck</h2>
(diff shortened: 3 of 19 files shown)


next/.babelrc
```diff
@@ -0,0 +1,3 @@
+{
+  "presets": ["next/babel"]
+}
```
lily-pytel: This file is needed for jest. It didn't seem to interfere with next but let me know if it does anything on Mac

next/components/card-list.js
```diff
@@ -6,6 +6,7 @@ export default function CardList({ cards }) {
     <ul className="cardList">
       {cards.map((card, index) => (
         <li key={card.id}>
+          {card.quantity && <span>{card.quantity}</span>}&nbsp;
```
lily-pytel: I'm reusing the card-list component for now to show the imported cards, just added a possible quantity.

next/components/imported-deck.js
```diff
@@ -0,0 +1,71 @@
+import PropTypes from 'prop-types';
```
lily-pytel: Component to show a list of import results, like so:

image

next/next.config.js
```diff
@@ -2,5 +2,12 @@ module.exports = {
   env: {
     GA_TRACKING_ID: 'UA-142178156-1',
     USE_GOOGLE_ANALYTICS: process.env.NODE_ENV === 'production'
+  },
```
lily-pytel: Added to get live-reloading on Windows 😭

next/pages/deckbuilder.js
```diff
@@ -0,0 +1,98 @@
+import React from 'react';
```
lily-pytel: Main page component for deck builder

image

next/components/imported-deck.js
```diff
@@ -0,0 +1,71 @@
...
+    sideboard: PropTypes.arrayOf(
+      PropTypes.shape({
+        id: PropTypes.string,
+        quantity: PropTypes.number,
+        name: PropTypes.string
+      })
+    ),
+    errors: PropTypes.arrayOf(PropTypes.string)
+  })
+};
```
alexmiddeleer: This component’s ui kinda just looks like a deck. Maybe add these changes to the deck component and consume it?

next/components/imported-deck.js
```diff
@@ -0,0 +1,71 @@
...
+      </div>
+    );
+  }
+
+  const sideboardElement =
+    sideboard && sideboard.length ? (
+      <React.Fragment>
+        <h3>Sideboard</h3>
+        <CardList cards={sideboard} />
+      </React.Fragment>
```
alexmiddeleer: You can omit React.Fragment and just use an empty tag afaik

next/lib/import-utils.js
```diff
@@ -0,0 +1,99 @@
...
+    .join(':')
+    .trim();
+};
+
+export const formatCardLines = cardLines => {
+  const formatted = cardLines
+    .filter(line => line && line.trim && line.trim())
+    .map(line => {
+      const trimmed = line && line.trim && line.trim();
+      const split = trimmed && trimmed.split && trimmed.split(' ');
```
alexmiddeleer: I don’t think you need to be this defensive - can cause errors to be more confusing imo. Not a big deal though

next/lib/import-utils.js
```diff
@@ -0,0 +1,99 @@
...
+        !split.length ||
+        split.length < 2 ||
+        isNaN(parseInt(split[0]))
+      ) {
+        return line;
+      }
+
+      return {
+        id: 'TBD',
+        quantity: parseInt(split[0]),
```
alexmiddeleer: It’s recommended to always supply a radix to parseint. I forget exactly the reason atm.

next/lib/import-utils.js
```diff
@@ -0,0 +1,99 @@
...
+      split.length < 2 ||
+      !line[0] ||
+      !line[0] === metaName ||
+      !line[1]
+  );
+};
+
+export const cardLinesValid = lines => {
+  return lines.reduce(
+    (acc, curr) => Boolean(acc && curr.quantity && curr.name),
```
alexmiddeleer: Nice 👌

next/pages/deckbuilder.js
```diff
@@ -0,0 +1,98 @@
...
+  const sideboardLines = sideboardText.split(/\n/g);
+
+  importedDeck.deckName = extractMetaValue(mainDeckLines[0]);
+  importedDeck.deckPath = extractMetaValue(mainDeckLines[1]);
+  importedDeck.deckPower = extractMetaValue(mainDeckLines[2]);
+  importedDeck.mainDeck = formatCardLines(mainDeckLines.slice(3));
+  importedDeck.sideboard = formatCardLines(sideboardLines);
+
+  return importedDeck;
+};
```
alexmiddeleer: Maybe this should live in import utils too

next/pages/deckbuilder.js
```diff
@@ -0,0 +1,98 @@
...
+    };
+
+    this.handleInputChange = handleInputChange.bind(this);
+    this.handleImport = this.handleImport.bind(this);
+  }
+
+  handleImport() {
+    const { mainDeckInput, sideboardInput } = this.state;
+    this.setState({
+      importedDeck: convertImportToDeck(mainDeckInput, sideboardInput)
```
alexmiddeleer: Are md  and sb separate when you export from the game? If not we might just want one import field, not two. Minimal effort for the user.

next/lib/import-utils.js
lily-pytel: @alexmiddeleer you sure? As you can see I'm really paranoid in Javascript, since errors just wipe out the whole website. That's also why I try to put nulls and junk in the tests to make sure things don't fail for nullability - but I can chill about that and take some of these out

next/lib/import-utils.js
lily-pytel: Ugh right, I forgot about that

next/components/imported-deck.js
lily-pytel: Nice, did not know that

next/pages/deckbuilder.js
lily-pytel: @alexmiddeleer it looks to me like the game doesn't actually have a sideboard so this is more of a fluff feature - it's why it's not included in the import as well. The sideboard is not included in the export of eternal warcry either. We can get rid of it alltogether

next/components/imported-deck.js
lily-pytel: @alexmiddeleer The `deck` component has a query and then uses `CardList` to show the list of cards, which this file is also using, so I think they are sharing the UI that they have in common, no?

alexmiddeleer: Yeah, I think you're right. Though I think we should still add the name, path and power to the deck component.

alexmiddeleer: Oops, and you have Path: twice in the gui instead of Path and Power.

next/lib/import-utils.js
alexmiddeleer: Curious what others think, but I never write code like this on the front-end. Angular and Ember keep going after a javascript error, does React just die?

jtrussell: Depends where the error happens. If you throw without catching during a render, and you don't have error boundaries set up, you'll get a white page of death. Setting up error boundaries is a decent idea if you believe you're likely to encounter them and want to handle them gracefully ("oops we messed up" graceful as opposed to "it's broken but here's something" graceful).

In this particular case, the checks for `trim` and `split` inside the `map` block are redundant. In the `filter` block we're already calling `trim` on each item - so `line.trim()` is safe on 27 and `line.trim().split()` (line 28) has to work unless someone got really mad and clobbered `String.prototype.trim` or `String.prototype.split`. Looking at our internal use, I only see `formatCardLines` getting called on the result of a `String.prototype.split` which should only give us arrays of things we could safely `trim()` and `split()`. I'd probably not be so defensive on an internal API but I would happily adopt something like TypeScript for the safe feels. Creating a `CardLines` class (or whatever) to force that structure is also a fine option.

From a style standpoint, I'd probably validate the input at the top of the function and go on with my life in YOLO mode... And if that isn't possible there's always `try ... catch ...` which some folks may find more to their liking.

next/next.config.js
jtrussell: 😭 I'm going to see if there's an easy way to opt out this for non-Windows folks 🤔.

next/package.json
```diff
@@ -25,11 +28,18 @@
     "react-markdown": "^4.1.0"
   },
   "devDependencies": {
+    "babel-core": "^6.26.3",
     "babel-eslint": "^10.0.2",
+    "babel-jest": "^24.8.0",
+    "babel-preset-env": "^1.7.0",
+    "babel-preset-react": "^6.24.1",
```
jtrussell: `babel-preset-env` and `babel-preset-react` might not be necessary - or at least the deps in the ["with-jest" next demo](https://github.com/zeit/next.js/blob/canary/examples/with-jest/package.json) doesn't include them.

next/pages/card.js
```diff
@@ -18,7 +18,7 @@ export const cardQuery = gql`
 `;
 
 export default withRouter(({ router }) => (
-  <Query query={cardQuery} variables={{ id: parseInt(router.query.id) }}>
+  <Query query={cardQuery} variables={{ id: parseInt(router.query.id, 10) }}>
```
jtrussell: Good on you 👍. Always provide a radix.

next/.babelrc
jtrussell: This is cool - it's actually just the babel preset used by Next in a neat little package. Won't change anything for Next, it's using this internally, and now Jest can be on the same page. For ref: https://github.com/zeit/next.js/blob/canary/packages/next/babel.js

next/components/imported-deck.js
lily-pytel: Agreed about adding it to the deck, we have a card for it here: https://trello.com/c/SQvNN74c/41-add-deck-name-power-and-path-to-the-deck-page

lily-pytel: Fixed typo!

next/lib/import-utils.js
lily-pytel: > has to work unless someone got really mad and clobbered String.prototype.trim or String.prototype.split

Why would you not assume that? 😛 

Anyway replaced some of the deep paranoia with try-catch. Hope that's better

next/package.json
lily-pytel: True! It's ded

PR #7, opened 2019-07-28
Feature/export deck

lily-pytel:
Builds off this PR: https://github.com/mythgard-hub/mythgard-hub/pull/6

- added export functionality
- in the deckbuilder page, people can export a deck once they're imported it successfully. It doesn't do the sideboard because is there even a sideboard in this game? I didn't notice it

image

- in the deck page, people can export the deck but it will add a temporary name, path and power. it will also add a default quantity of 1 to each card. we'll need to change the format of the deck page to have the user be able to change those things

image

### DIFF
--- a/next/components/deck-export.js
+++ b/next/components/deck-export.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import { copyToClipboard } from '../lib/copy-to-clipboard';
 
 class DeckExport extends React.Component {
   constructor(props) {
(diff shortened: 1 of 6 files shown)


next/components/deck-export.js
```diff
@@ -0,0 +1,49 @@
...
+      return;
+    }
+
+    const success = copyToClipboard(textToExport);
+    this.setState({ message: success ? 'Copied' : 'Failed to copy' });
+
+    // Make message disappear
+    window.setTimeout(() => {
+      this.setState({
+        message: null
```
alexmiddeleer: Can we make a disappearing  message component? Could just call it a toast I suppose.

lily-pytel: @alexmiddeleer yeah, but I didn't think we were getting into styles and stuff quite yet. Do you want me to make a trello card for it?

alexmiddeleer: I was just thinking, as a quick refactor, that it might be nice to extract this code into a separate component.

PR #8, opened 2019-07-30
Tournaments

evsheffield:
A very simple skeleton for the tournament pages. I don't think there's anything especially worth calling out, but of course I'm happy to hear any comments. Granted this is p simple right now but so far getting a free GraphQL API without having to write rest endpoints feels like the best thing ever.

### DIFF
--- a/e2e/cypress/integration/tournaments-page.js
+++ b/e2e/cypress/integration/tournaments-page.js
@@ -3,11 +3,14 @@ describe('Tournaments Page', function() {
     cy.visit('/tournaments');
   });
   it('should have a list of clickable tournaments', function() {
-    cy.get('ul.tournamentList li').should('have.length', 2);
+    cy.get('[data-cy=tournamentList] [data-cy=tournamentListItem]').should(
+      'have.length',
+      2
+    );
 
-    cy.get('ul.tournamentList a:first').click();
+    cy.get('[data-cy=tournamentList] a:first').click();
     cy.location().should(location => {
       expect(location.pathname).to.eq('/tournament');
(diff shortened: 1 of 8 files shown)


next/components/tournament-list.js
```diff
@@ -0,0 +1,27 @@
+import Link from 'next/link';
+import PropTypes from 'prop-types';
+
+TournamentList.propTypes = {
+  tournaments: PropTypes.arrayOf(
+    PropTypes.shape({
+      id: PropTypes.number.required,
+      name: PropTypes.string.required
```
jtrussell: Nit pick - I saw we had `isRequired` below vs `required` here. They both work but bonus points shall be awarded for consistency :).

evsheffield: Ah oopsie! I usually go for `isRequired` so I'll get those tidied up.

e2e/cypress/integration/tournaments-page.js
```diff
@@ -0,0 +1,13 @@
+describe('Tournaments Page', function() {
+  beforeEach(() => {
+    cy.visit('/tournaments');
+  });
+  it('should have a list of clickable tournaments', function() {
+    cy.get('ul.tournamentList li').should('have.length', 2);
```
alexmiddeleer: My earlier examples didn't follow this, my apologies (I'm still learning these too). But let's try to use selectors according to cy best practices, e.g. `data-cy="tournamentListItem"` (https://docs.cypress.io/guides/references/best-practices.html#Selecting-Elements)

PR #9, opened 2019-07-31
Add Google Auth

jtrussell:
There are still plenty of things I'm not happy with here but this definitely has potential to sprawl and I feel like I've already had the branch open for a bit. I'm opening the PR to get some feedback on what you all think makes sense to implement before merging and what can be left as TODOs for the future.

Some notes:
- You'll need a `.env` file in the repo root the run the project with docker-compose. There's a `.env.template` with some placeholders. `.env` itself won't be tracked by git. I'll share the necessary secrets to get auth running. All non-auth things should run if you just rename `.env.template` to `.env` and use the placeholders). If we like this approach (rather than say, using `dotenv` and loading things in-app) we can migrate other secrets still hiding in code to this file.
- To track auth I added session support to the express app that wraps our next server. I actually started out with all this in the express service and then realized I'd have to proxy requests to the express service whenever we wanted auth things in next land... and that didn't sound like much fun. It would probably be better to use cookie based JWTs... I'll be honest I got tired and used session storage because that's what I know 😅.
- I feel like the stored function to find/create users by google id could be better. I'm pretty sure PostgreSQL has a way to return a value you just inserted/updated without needed a subsequent select... for whatever reason that nugget of knowledge is hiding in corners of Google that I'm overlooking. Oh - also I need to check what happens with conflict resolution when your Google ID is null... I'm pretty sure PSQL treats null as not equal to null 🤔.
- I was hoping my `withUser` HOC could be applied to the `_app.js` page template. Still working on making that happen but right now it should be applied to individual pages as in `pages/privacy-policy`. Yeah, Google asks you to register a privacy policy page when you set up your OAuth credentials.
- Oh, also - there's no "pre-login-url" handling going on.


jtrussell:
**Don't** be gentle 😅

evsheffield:
RE returning the result of the manipulated row in the stored function, is `RETURNING` perhaps what you're looking for? https://www.postgresql.org/docs/9.3/dml-returning.html

jtrussell:
I had tried `RETURNING` but couldn't get it to work in the context of the stored function... though I was far from exhaustive. It's definitely worth playing around with a bit more.

alexmiddeleer:
This looks really promising. I like the idea of an auth HOC, and it looks like everything is pretty neat and tidy. Does this handle users trying to send unauthorized graphql requests, or do we still need to figure that out? Postgraphile has some recommendations for auth but they're likely not trivial and we may not want them. I'm okay with creating a whitelist of legal mutations if that's the easiest way.

jtrussell:
We still need to figure out users trying to send unauthorized graphql requests. I suppose we at least will get a user reference attached to the requests though.

jtrussell:
Good call on checking posgraphile, @alexmiddeleer. Looks like they really want you in the JWT camp, guess I'll learn me how JWTs work 🤓. What I'll probably do is get the rest of the surrounding APIs how I'd like it (see below), merge this, then work in changing the session tracking for JWTs. At that point, securing the graphql endpoints can probably fall under a separate PR too.

After sitting with this for a couple days I think I'm also going to change the `withAuth` HOC to a provider/consumer model. As it is right now there isn't a great place to toss `withAuth` in `_app.js` and you really want it to run on every page so you can be sure it runs in the `getInitialProps` on the SSR page. You need a server render to get the auth info. Alternatively, we could make a request to the server for auth status but that's annoying too.

Good call on `RETURNING` @evsheffield - I had some other junk that was flaking out and mis-attributed the no-working-ness to `RETURNING`.

jtrussell:
Now with JWTs instead of sessions. I'm still figuring a couple things outs with JWTs but this should be good enough to merge and not blog things that require auth any longer. I'll give folks another day or so to review the changes in case anything makes you 🤢.

next/server-routes/auth.js
```diff
@@ -0,0 +1,108 @@
...
+});
+
+router.get(
+  '/google',
+  passport.authenticate('google', {
+    scope: process.env.GOOGLE_AUTH_SCOPE,
+    session: false
+  }),
+  (req, res, next) => {
+    console.log('WAT?!');
```
alexmiddeleer: ✂️ ?

jtrussell: 👍

jtrussell:
Thanks! I added a note to the README indicating what we're using for auth including links to relevant READMEs. To be honest the flow follows a standard Passport implementation - I'm not sure I could explain it better than their README.

JWTs aren't super magical. They're essentially JSON strings (or other strings) that the server encrypts before handing to the client, then decrypts to verify contents. The idea being that, as long as your encryption key is not compromised, the encrypted JSON can be treated as a receipt for auth status, or whatever else you might want a receipt for. [Like buying donuts](https://www.youtube.com/watch?v=gWx6uA5aCrE). Typically you have an expiration baked into each token. If you're paranoid you might also rotate your encryption keys regularly and use secure or http only cookies. From what I could see cookies seem to be the way the cool kids are storing JWTs these days, which makes it pretty easy to get them back to the server. This only gets a little awkward in `getInitialProps` where something like `unfetch` or `fetch` from a non-browser either won't send cookies or won't have cookies to send.

alexmiddeleer:
@jtrussell Thanks again for getting this done - never a quick and easy step for an app.

jtrussell:
@alexmiddeleer Mostly awkward because Next _really_ wants to be stateless 😅.

PR #10, opened 2019-08-01
Deck builder

alexmiddeleer:
Right now the deck importer and exporter does not work at all with the new card picking and deck saving logic. I added a Trello card for this.


next/components/deck-card-list.js
```diff
@@ -0,0 +1,21 @@
...
+import PropTypes from 'prop-types';
+
+export default function DeckCardList({ cards }) {
+  return (
+    <ul className="deckCardList" data-cy="deckCardList">
+      {cards.map((card, index) => (
+        <li key={index}>
+          <Link href={`/card?id=${card.id}`} key={index}>
+            <a data-cy="deckCardListCard">
+              {card.id}. name: {card.name}
```
lily-pytel: Got rid of the quantity?

lily-pytel: I see Evan is doing the card to add quantities to a deck so it's probably not relevant here

next/components/card-list.js
```diff
@@ -1,14 +1,19 @@
...
+    <ul className="cardList" data-cy="cardList">
       {cards.map((card, index) => (
         <li key={card.id}>
           {card.quantity && <span>{card.quantity}</span>}&nbsp;
           <Link href={`/card?id=${card.id}`} key={index}>
-            <a>
+            <a
+              data-cy="cardListCard"
+              onClick={e => {
+                onCardClick && onCardClick(e, card);
```
evsheffield: Not a huge difference but if you wanted to avoid checking if this prop exists you could define its default as a no-op function. I find myself doing this a lot for props that are event handlers.

```
CardList.defaultProps = {
  onCardClick: () => {}
}
```

next/pages/deck-builder.js
```diff
@@ -0,0 +1,200 @@
...
+
+  handleImport() {
+    const { mainDeckInput, sideboardInput } = this.state;
+    this.setState({
+      importedDeck: convertImportToDeck(mainDeckInput, sideboardInput)
+    });
+  }
+
+  validateState() {
+    if (!this.state.deckName) {
```
evsheffield: I'm sure we're expecting this function to change but for now it could be simplified to `return !!this.state.deckName`.

alexmiddeleer: Good feedback, but I did write it anticipating more validations. Maybe a YAGNI violation but I'm gonna leave it in for now.

alexmiddeleer:
Oops, apologies for forgetting to squash. Bad habit.

PR #13, opened 2019-08-02
Deck builder connect import

lily-pytel:
- The deck name gets updated when an import occurs
- The current deck gets updated when the import occurs
- The sideboard has been hidden but in a way that doesn't break any tests

TODO
- When the deck is saved, the imported decks are ignored because they all have ID 'TBD' instead of an int (yay? but also fix)
- When the deck is exported, it only exports the imported cards (there is a trello card for that)
- No new e2e tests (will happen in the next trello)
- The import and deck builder have not been connected in the most graceful of ways. Specifically we have a state with a nest object which I'm not a fan of (although is quite convenient). Could use some love but perhaps a bit pointless until we decide on a UI. Could change things.


next/lib/import-utils.js
```diff
@@ -1,10 +1,11 @@
 import { META_KEYS } from '../constants/deck';
 
-export const initializeImportedDeck = () => {
+export const initializeDeckBuilder = () => {
   return {
     deckName: '',
     deckPath: '',
     deckPower: '',
+    deckCoverart: '',
```
alexmiddeleer: `deckCoverArt`? Not sure but looking wrongly camelized

next/pages/deck-builder.js
```diff
@@ -92,36 +91,44 @@ class DeckBuilderPage extends React.Component {
...
   onCollectionClick(e, card) {
+    const { deckInProgress } = this.state;
     e && e.preventDefault();
+
     this.setState({
-      deckInProgress: [...this.state.deckInProgress, card]
+      deckInProgress: {
+        ...deckInProgress,
+        mainDeck: [...deckInProgress.mainDeck, card]
+      }
```
alexmiddeleer: Fancy! I assume the mainDeck "wins" over the first `...deckInProgress` which may have its own `mainDeck` prop already. It's deckInProgress inception. But seems fine, I don't have a problem with it aside from it looking a bit weird at first.

next/pages/deck-builder.js
```diff
@@ -92,36 +91,44 @@ class DeckBuilderPage extends React.Component {
...
   }
 
   handleImport() {
-    const { mainDeckInput, sideboardInput } = this.state;
+    const { mainDeckInput, deckInProgress } = this.state;
+
+    const importedDeck = convertImportToDeck(mainDeckInput, '');
+    importedDeck.mainDeck = deckInProgress.mainDeck.concat(
+      importedDeck.mainDeck
+    );
```
alexmiddeleer: could use the fancy `importedDeck.mainDeck = [...deckInProgress.mainDeck, ...importedDeck.mainDeck]` right? Or am I missing something? It's less efficient but efficiency doesn't matter here.

alexmiddeleer: Hmm, should the mainDeck be replaced, not appended to? Maybe a question for Mark and co. before you change anything.

next/pages/deck-builder.js
```diff
@@ -144,8 +151,15 @@ class DeckBuilderPage extends React.Component {
...
                   type="text"
                   name="deckName"
-                  onChange={this.handleInputChange}
-                  value={this.state.deckName}
+                  onChange={e => {
+                    this.setState({
+                      deckInProgress: {
+                        ...deckInProgress,
+                        deckName: e.target.value
+                      }
```
alexmiddeleer: This feels a bit awkward. Maybe there's no terse way since we need to use setState, but extracting this to a helper like `onChange={updateDeckName}` might feel cleaner. Anyone else have a better idea?

next/lib/import-utils.js
```diff
@@ -1,10 +1,11 @@
 import { META_KEYS } from '../constants/deck';
 
-export const initializeImportedDeck = () => {
+export const initializeDeckBuilder = () => {
```
alexmiddeleer: is this still living in a reasonable spot or should it live in `deck-builder.js`?

next/lib/import-utils.js
lily-pytel: Will do in the next card which is to add this to the import. I still need to find out what it looks like since the imports I've done have not had this. Looks like it's an optional thing but it's not an empty line like power or path which are also options. SMH.

next/pages/deck-builder.js
lily-pytel: Gotta love Javascript's `...` magic. The existing `mainDeck` will be smashed by the new one. I think this could do with some refactoring actually, so we don't have an object in the state... but later. Maybe.

next/pages/deck-builder.js
lily-pytel: Fun fact: In the case of two arrays, this will happen:
```
const arr = [1, 2, 3]
const arr2 = [4, 5, 6]
console.log([...arr, arr2])
-- [1, 2, 3, [4, 5, 6]]
```

In other words, the `...` appends to the end of the array but in this case we do want to concat the two. 

As for your other question, in eternal warcry the import does replace anything you had before (seems a little bad to me), so yeah, should confirm.

next/pages/deck-builder.js
lily-pytel: Probably to use formik but I put it in a helper function for now so it looks better.

next/lib/import-utils.js
lily-pytel: It is used in this file as well. It can live in its own util file called deck-utils.js or something

lily-pytel: Done

next/pages/deck-builder.js
jtrussell: Yeah, pretty common pattern for updating state in a redux reducers.

```
nextState = { ...oldState, updateKey: updateValue }
```

next/pages/deck-builder.js
jtrussell: `[...arr, ...arr2]` works if you spread the second array too... but then there will be fewer cats in your code.

lily-pytel: @jtrussell oh yeah, I didn't of that! 🤦‍♂

alexmiddeleer: Oops, yeah that was the syntax I meant to recommend. I always forget the second set of `...`.

PR #18, opened 2019-08-04
Add cover art to deck import + loosen the rules of input to match the game more

lily-pytel:
Trello: https://trello.com/c/M4rOTeVM/51-deckbuilder-hide-sideboard-and-add-coverart-to-import-export

- Cover art was added to the constants and default deck in progress
- The import was changed so that it accepts any number of meta values - looks like the game is more flexible and doesn't actually require a name, path, power or anything
- Tests were changed to reflect that part

TODO

- export will be fixed in another PR
- e2e tests will be added in another PR

### DIFF
--- a/next/lib/import-utils.js
+++ b/next/lib/import-utils.js
@@ -1,15 +1,48 @@
 import { META_KEYS } from '../constants/deck';
 import { initializeDeckBuilder } from './deck-utils';
 
-export const extractMetaValue = line => {
+const hasCoverArt = line => {
+  return line.startsWith(`${META_KEYS.COVER_ART}:`);
+};
+
+// Get the line number where the meta information ends
+export const getSpliceIndex = lines => {
+  let index = 0;
+
+  while (index < lines.length) {
+    const currSplit = lines[index].split(':');
+
+    if (currSplit.length === 1) {
+      return index;
+    }
+
+    index++;
+  }
+
+  return index;
+};
+
+// Try to find a meta value. There could be 0-4 meta values
+export const extractMetaValue = (lines, metaname) => {
+  if (!Array.isArray(lines)) {
+    return null;
+  }
+
   try {
-    return line
-      .split(':')
-      .slice(1)
-      .join(':')
-      .trim();
+    let index = 0;
+
+    while (index < lines.length) {
+      const currSplit = lines[index].split(':');
+
+      if (currSplit[0] === metaname && currSplit[1]) {
+        return currSplit[1].trim();
+      }
+
+      index++;
+    }
+
+    return '';
   } catch (e) {
-    console.log(e);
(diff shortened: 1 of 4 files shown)


next/lib/import-utils.js
```diff
@@ -1,15 +1,44 @@
...
 
-export const extractMetaValue = line => {
+// Get the line number where the meta information ends
+export const getSpliceIndex = lines => {
+  let index = 0;
+
+  while (index < lines.length) {
+    const currSplit = lines[index].split(':');
+
+    if (currSplit.length === 1) {
```
evsheffield: If we're just going on the presence of a colon to indicate meta lines perhaps a quicker check would be `lines[index].indexOf(':') === -1`.

next/lib/import-utils.js
```diff
@@ -45,18 +73,19 @@ export const formatCardLines = cardLines => {
 };
 
 export const metaLineInvalid = (line, metaName) => {
+  // Gotta have the basics
+  if (!metaName || !line || !line.split) {
+    return true;
+  }
+
   const split = line && line.split && line.split(':');
```
evsheffield: I think now that we have the check above we no longer need to check `line && line.split` since they must have been true to reach here.

next/lib/import-utils.js
```diff
@@ -45,18 +73,19 @@ export const formatCardLines = cardLines => {
...
-      !line ||
-      !split ||
-      !split.length ||
-      split.length < 2 ||
-      !line[0] ||
-      !line[0] === metaName ||
-      !line[1]
-  );
+  // The first element of the line has to be the meta name
+  if (!split || !split.length || split[0] !== metaName || split[0] === line) {
```
evsheffield: Is there any situation in which `!split` will return a truthy value? So hard to keep track of this type of stuff in JS but I think we should always end up with an array result from `.split()` so the first check might be redundant.

evsheffield: We should be able to just return the result of this expression instead of using the if/else.

next/lib/import-utils.js
lily-pytel: Oh yeah, true

next/lib/import-utils.js
lily-pytel: You're right - we were checking way more things before but now it just comes down to this expression

lily-pytel: The rules of import are apparently much more chill than I originally thought

lily-pytel: If there ever is, we'll catch it. I don't see the harm in doing this

next/lib/import-utils.js
lily-pytel: Truth

PR #19, opened 2019-08-04
Card quantities

evsheffield:
I know we just went through the trouble of removing, but I ended up restoring `deck-card-list`. The data structure that each takes is slightly different (i.e. cardDeck vs. a plain ol' card), and beginning with quantity I expect that card list may diverge in appearance/behavior depending on whether they are displayed in the context of a deck or just as a plain list of cards. If you guys have thoughts otherwise I'm happy to make adjustments.

I think this will cause some conflicts with Lily's PR since I've modified the main card list to be an object for easier lookup; I can clean those up once that gets merged.

I've called out as a TODO some validation around imported cards, since in order to properly save them to the deck we will need to associate them with their IDs. I'll add this as a separate card since it seems like we have some options for how to approach this.

### DIFF
--- a/next/components/card-list-item.js
+++ b/next/components/card-list-item.js
@@ -1,0 +1,25 @@
+import React from 'react';
+import Link from 'next/link';
+import PropTypes from 'prop-types';
+
+export default function CardListItem({ card, onCardClick }) {
+  return (
+    <Link href={`/card?id=${card.id}`}>
+      <a data-cy="cardListCard" onClick={e => onCardClick(e, card)}>
+        {card.id}. name: {card.name}
+      </a>
+    </Link>
+  );
+}
+
+CardListItem.propTypes = {
+  card: PropTypes.shape({
+    id: PropTypes.number.isRequired,
+    name: PropTypes.string.isRequired
+  }).isRequired,
+  onCardClick: PropTypes.func
+};
+
+CardListItem.defaultProps = {
+  onCardClick: () => {}
+};
--- a/next/components/card-list-item.js
+++ b/next/components/card-list-item.js
@@ -2,10 +2,10 @@ import React from 'react';
 import Link from 'next/link';
 import PropTypes from 'prop-types';
 
-export default function CardListItem({ card, onCardClick }) {
+export default function CardListItem({ card, onClick }) {
   return (
     <Link href={`/card?id=${card.id}`}>
-      <a data-cy="cardListCard" onClick={e => onCardClick(e, card)}>
+      <a data-cy="cardListCard" onClick={e => onClick(e, card)}>
         {card.id}. name: {card.name}
       </a>
     </Link>
@@ -17,9 +17,9 @@ CardListItem.propTypes = {
     id: PropTypes.number.isRequired,
     name: PropTypes.string.isRequired
   }).isRequired,
-  onCardClick: PropTypes.func
+  onClick: PropTypes.func
 };
 
 CardListItem.defaultProps = {
-  onCardClick: () => {}
+  onClick: () => {}
 };
--- a/next/pages/deck-builder.js
+++ b/next/pages/deck-builder.js
@@ -11,6 +11,7 @@ import Router from 'next/router';
 import DeckExport from '../components/deck-export';
 import { convertImportToDeck } from '../lib/import-utils';
 import { initializeDeckBuilder } from '../lib/deck-utils';
+import DeckCardList from '../components/deck-card-list';
 
 const addDeckQuery = gql`
   mutation AddDeck($name: String!) {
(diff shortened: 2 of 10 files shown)


evsheffield:
NVM that TODO as it looks like Lily is already on it. https://trello.com/c/AZUleTYz/52-deckbuilder-check-imported-cards-are-valid-before-adding-to-deck

lily-pytel:
All merged now, if you wanna settle those conflicts. I see `mainDeck` is an object now?

next/components/card-list-item.js
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+import Link from 'next/link';
+import PropTypes from 'prop-types';
+
+export default function CardListItem({ card, onClick }) {
+  return (
+    <Link href={`/card?id=${card.id}`}>
+      <a data-cy="cardListCard" onClick={e => onClick(e, card)}>
```
jtrussell: FWIW having an `a` tag, especially one with an `href` prop, that prevents the default event behavior and then does JavaScripty things is p un-a11y friendly. I know this PR didn't add the functionality, just calling it out as something to keep in mind if we can avoid doing it in the future.

Not sure how many folks this would affect are playing MG... but hey 😅.

next/pages/deck-builder.js
```diff
@@ -24,9 +25,14 @@ const addDeckQuery = gql`
 `;
 
 const addCardDeck = gql`
-  mutation CreateCardDeck($deckId: Int!, $cardId: Int!) {
-    createCardDeck(input: { cardDeck: { deckId: $deckId, cardId: $cardId } }) {
+  mutation CreateCardDeck($deckId: Int!, $cardId: Int!, $quantity: Int!) {
```
jtrussell: Neat - I hadn't seen a mutation in the wild before. Is this a standard graphql thing or a an apollo-ism?

postgres/init.sql
```diff
@@ -23,6 +23,7 @@ INSERT INTO mythgard.deck("name") VALUES ('cats');
 
 CREATE TABLE mythgard.card_deck (
   id SERIAL PRIMARY KEY,
+  quantity integer,
```
alexmiddeleer: Interesting, I am not sure I would have done it this way. Did you consider calculating the quantity on the front-end by counting up the number of card_deck records for deck_id X that have card_id Y? It might help prevent weird bugs if another card_deck record gets somehow inserted into the table with the same card_id and deck_id. I'm certainly not against this approach if you think it's a good one, but I wouldn't mind chatting about it or hearing what others think. I am by no means a databases expert - far from it.

next/pages/deck-builder.js
alexmiddeleer: standard graphql

postgres/init.sql
lily-pytel: I would have also thought to do it with a `quantity` field myself, as opposed to having a separate record for each card like a real deck - however, that might help us if there are ever any sets and we need to specify which copy from which set for some reason

next/components/card-list-item.js
alexmiddeleer: We may just want to do something funky where we have two totally different templates if onClick is provided. I do think a11y is important. Good call.

evsheffield: It's definitely a good call. Since you brought it up I did observe (and promptly forget) how it's strange that when you hover over it your browser displays the link in the corner but then the app is like nah bro. At any rate I've added a new card to our backlog to clean that up: https://trello.com/c/gWt3CAg9/61-clean-up-a-tags-where-behavior-overrides-links

postgres/init.sql
evsheffield: I did consider that as it does seem simpler in some ways, but went with this approach for a few reasons:

- I think it keeps the data representation closer to the GUI. We often see quantity right next to the card so it made sense to me to keep those together.
- Not that it's necessarily an issue but it's overall fewer records in the DB. Helps deal with clowns who want to put 9001 shiny Charizards in their deck.
- It avoids having to do any list traversals to calculate quantities. Again this may not be an issue in practice but with this structure getting card quantity for any given card is `O(1)` which feels pretty good.

To your point of getting duplicate `card_id`/`deck_id` combos, I think we should be able to prevent this with a DB constraint. I've done it with other SQL DBs, just not postgres but I'll try to add something to take care of that. 👍 

I'm not sure if the landscape of digital card games leads to "sets", but I think if we ever did end up in that situation that it would be perfectly reasonable to have identical cards from different sets represented as unique cards in the DB. I think either DB structure should be able to support that case if it arises.

evsheffield: I did consider that as it does seem simpler in some ways, but went with this approach for a few reasons:

- I think it keeps the data representation closer to the GUI. We often see quantity right next to the card so it made sense to me to keep those together.
- Not that it's necessarily an issue but it's overall fewer records in the DB. Helps deal with clowns who want to put 9001 shiny Charizards in their deck.
- It avoids having to do any list traversals to calculate quantities. Again this may not be an issue in practice but with this structure getting card quantity for any given card is `O(1)` which feels pretty good.

To your point of getting duplicate `card_id`/`deck_id` combos, I think we should be able to prevent this with a DB constraint. I've done it with other SQL DBs, just not postgres but I'll try to add something to take care of that. 👍 

I'm not sure if the landscape of digital card games leads to "sets", but I think if we ever did end up in that situation that it would be perfectly reasonable to have identical cards from different sets represented as unique cards in the DB. I think either DB structure should be able to support that case if it arises.

alexmiddeleer: Ok, that's reasonable. Thanks for the explanation 👍

next/components/card-list-item.js
alexmiddeleer: perfect

PR #20, opened 2019-08-08
Export deck in progress

lily-pytel:
- Export was modified to export the deck in progress, not just the imported decks
- Unit tests added for converting the deck in progress to a string that can be added to clipboard
- Added fairly lame e2e tests for export and import. Seems like you can't test the content of the clipboard in cypress but hopefully the unit tests cover that it's the right stuff. As for import, I think this test is going to be changed HEAVILY once we have some styles and such, so it seems pointless to make it super intricate.

### DIFF
--- a/next/components/deck-export.js
+++ b/next/components/deck-export.js
@@ -53,9 +53,10 @@ DeckExport.propTypes = {
     deckPower: PropTypes.string,
     deckCoverArt: PropTypes.string,
     mainDeck: PropTypes.shape({
-      id: PropTypes.number,
       quantity: PropTypes.number,
-      name: PropTypes.string
+      card: PropTypes.shape({
+        name: PropTypes.string
+      })
     }),
     errors: PropTypes.arrayOf(PropTypes.string)
   })
--- a/next/components/deck-export.js
+++ b/next/components/deck-export.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { copyToClipboard } from '../lib/copy-to-clipboard';
+import { exportDeck } from '../lib/export-utils';
 
 class DeckExport extends React.Component {
   constructor(props) {
@@ -11,7 +12,8 @@ class DeckExport extends React.Component {
   }
 
   handleExport() {
-    const { textToExport } = this.props;
+    const { deckInProgress } = this.props;
+    const textToExport = exportDeck(deckInProgress);
 
     if (!textToExport.trim()) {
       this.setState({ message: 'No imported deck to export' });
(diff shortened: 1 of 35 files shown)


alexmiddeleer:
👏

alexmiddeleer:
Oops, I reviewed this on my phone and didn’t see all the files. One minute

next/lib/export-utils.js
```diff
@@ -0,0 +1,24 @@
+export const exportDeck = deckInProgress => {
+  try {
+    const exportMeta = [
+      `name: ${deckInProgress.deckName}`,
+      `path: ${deckInProgress.deckPath}`,
+      `power: ${deckInProgress.deckPower}`,
+      `coverart: ${deckInProgress.deckCoverArt}`
+    ].join('\n');
```
alexmiddeleer: If you wanna be fancy, es6 templates support multiline out of the box. Completely unnecessary to change though. Also causes weird indentation in the file.

next/lib/export-utils.js
```diff
@@ -0,0 +1,24 @@
...
+
+    const cards = Object.values(deckInProgress.mainDeck)
+      .map(c => {
+        const quantity = c.quantity || '1';
+        const name = (c && c.card && c.card.name) || '';
+        return `${quantity} ${name}`;
+      })
+      .join('\n');
```
alexmiddeleer: Could use reduce instead of map. Nbd though.

next/lib/export-utils.test.js
```diff
@@ -0,0 +1,117 @@
+import { exportDeck } from './export-utils';
+import { initializeDeckBuilder } from './deck-utils';
+
+describe('Import utility methods', () => {
+  describe('Test exportDeck', () => {
+    it('No meta values', function() {
+      const deckInProgress = initializeDeckBuilder();
+      deckInProgress.mainDeck = [
```
evsheffield: I think this works because `Object.values` ends up giving you back an array with the same values (TIL!), but how its being passed from the deck builder is as an object with integer ID keys and values being the objects you've got here.

next/components/deck-export.js
```diff
@@ -34,16 +36,29 @@ class DeckExport extends React.Component {
...
 }
 
 DeckExport.propTypes = {
-  textToExport: PropTypes.string
+  deckInProgress: PropTypes.shape({
+    deckName: PropTypes.string,
+    deckPath: PropTypes.string,
+    deckPower: PropTypes.string,
+    deckCoverArt: PropTypes.string,
+    mainDeck: PropTypes.shape({
```
evsheffield: I think this is slightly different than the structure of main deck which currently looks like `{ 1: { quantity: 2, card: { name: "Black Lotus" } } }`. I think you can use `PropTypes.objectOf` to make assertions about what the values in an object look like, but I'm not sure about keys. Maybe that would require a custom prop validator, so I'm not sure whether or not you think that effort is worth it. At least the assertion on the shape of the values I think could be useful since this ends up getting consumed with `Object.values` anyways.

lily-pytel: So no id on the card, huh? Just the name for now? Will fix

lily-pytel: Ah, no, it does have ID

PR #21, opened 2019-08-09
Make postgres/init.sql idempotent

jtrussell:
This also allows us to:

- Run the init script over and over without errors or different results.
- Easily get a clean DB slate without rebuilding the container.


jtrussell:
Sure thing - added `scripts/hotswap-db.sh` 👍.

PR #22, opened 2019-08-10
Add power and path to the deck page

evsheffield:


next/components/deck.js
```diff
@@ -39,16 +39,22 @@ const deckToExportText = (deckCards, deckName) => {
...
   return (
     <Query query={deckCardsQuery} variables={{ id: deck.id }}>
-      {({ loading, error, data: { deck } }) => {
+      {({ loading, error, data }) => {
         if (error) return <ErrorMessage message="Error loading decks." />;
         if (loading) return <div>Loading</div>;
 
-        const cards = deck.cardDecks.nodes;
-        const deckToExport = deckToExportText(cards, deck.name);
+        const cards = data.deck.cardDecks.nodes;
```
lily-pytel: These dots scare me so

jtrussell: If `data.deck.cardDecks.nodes` isn't an array the query should error out, right?

postgres/init.sql
```diff
@@ -15,13 +15,31 @@ INSERT INTO mythgard.card ("id", "name", "rules", "type", "atk", "def") VALUES (
 INSERT INTO mythgard.card ("id", "name", "rules", "type", "atk", "def") VALUES (3, 'Grizzly Bear', 'rules', 'bear', '2', '2');
 INSERT INTO mythgard.card ("id", "name", "rules", "type", "atk", "def") VALUES (4, 'Dragon', 'flying', 'dragon', '5', '5');
 
+DROP TABLE IF EXISTS mythgard.path;
```
jtrussell: <3. Just FYI though the init script starts by killing everything in the schema. I'm not sure if that's better or worse than adding `drop table if exists` before each create. I guess this way would leave you with snippets that are easier to run by copy+paste when you just want to refresh a single table 🤔.

next/components/deck.js
lily-pytel: I don't think it would necessarily, or did I miss somewhere in the code that said either of those would never be null?

lily-pytel: Like is there some fancy validation done in postgraphql?

jtrussell: Hmm... I'm not sure about validating that it's an array (maybe there's an in inferred type coming in from postgraphile), but my understanding of vanilla graphql (which is limited) was that the query would error out if a response couldn't be put together that conformed to the requested schema.

lily-pytel: Yeah, but nulls are okay unless stated otherwise with the `!` all over the place. You saw from my previous PRs that I basically never just write anything with more than one `.` without validation or try-catch. Paranoid.

postgres/init.sql
alexmiddeleer: Yeah we can probably kill this line. If peeps are trying to keep their docker DB data over long periods it's probably a process smell.

postgres/init.sql
```diff
@@ -15,13 +15,31 @@ INSERT INTO mythgard.card ("id", "name", "rules", "type", "atk", "def") VALUES (
...
+
+INSERT INTO mythgard.path ("id", "name") VALUES (1, 'Way of the Black Lotus');
+
+DROP TABLE IF EXISTS mythgard.power;
+CREATE TABLE mythgard.power (
+  id SERIAL PRIMARY KEY,
+  name varchar(255)
+);
+
+INSERT INTO mythgard.power ("id", "name") VALUES (1, 'It''s over 9000!!');
```
alexmiddeleer: 😆

jtrussell: 💥  💇‍♂ 💥

next/components/deck.js
evsheffield: Interesting, clearly I have a lot to learn about how this all works. I can see the case where `data.deck` ends up being null, but if we do get back a deck are we not guaranteed to have `cardDeck.nodes` as an array? In my poking at it it seemed like the `cardDeck` connection was always there and `.nodes` was always an array, empty if there were no cards in the deck.

PR #23, opened 2019-08-10
Deck builder faction filter

alexmiddeleer:


next/components/some-cards.js
```diff
@@ -0,0 +1,89 @@
...
+          name
+          id
+        }
+      }
+    }
+  `;
+};
+
+class SomeCards extends Component {
+  static propTypes = {
```
lily-pytel: This is different than how we've been doing propTypes (Including in `FactionFilters` in this PR) - trying out something new?

alexmiddeleer: Trying a new autocompleter and this is how it adds proptypes 😄 . Good catch though, it's silly to be inconsistent.

alexmiddeleer: Going to fix this in the other PR because lazy.

PR #24, opened 2019-08-11
Deck card search

alexmiddeleer:
https://trello.com/c/tyKiqq1J/57-deck-seach-search-by-card-name-in-deck

### DIFF
--- a/next/components/card-search.js
+++ b/next/components/card-search.js
@@ -123,7 +123,7 @@ CardSearch.propTypes = {
 
 CardSearch.defaultProps = {
   cards: [],
-  onSelect: () => true
+  onSelect: () => {}
 };
 
 export default CardSearch;
--- a/next/components/card-search.js
+++ b/next/components/card-search.js
@@ -9,9 +9,7 @@ const getSuggestions = (value, cards) => {
 
   return inputLength === 0
     ? []
-    : cards.filter(
-        card => card.name.toLowerCase().slice(0, inputLength) === inputValue
-      );
+    : cards.filter(card => card.name.toLowerCase().indexOf(inputValue) > -1);
 };
 
 const getSuggestionValue = suggestion => suggestion.name;
--- a/next/components/card-search.js
+++ b/next/components/card-search.js
@@ -17,20 +17,6 @@ const getSuggestionValue = suggestion => suggestion.name;
 const renderSuggestion = suggestion => <div>{suggestion.name}</div>;
 
 class CardSearch extends Component {
-  static propTypes = {
-    cards: PropTypes.arrayOf(
-      PropTypes.shape({
-        id: PropTypes.number,
-        name: PropTypes.string
-      })
-    ),
-    onSelect: PropTypes.func
-  };
-  static defaultProps = {
-    cards: [],
-    onSelect: () => true
-  };
-
   constructor(props) {
     super(props);
     this.onChange = this.onChange.bind(this);
@@ -124,5 +110,20 @@ class CardSearch extends Component {
     );
   }
 }
+
+CardSearch.propTypes = {
+  cards: PropTypes.arrayOf(
+    PropTypes.shape({
+      id: PropTypes.number,
+      name: PropTypes.string
+    })
+  ),
+  onSelect: PropTypes.func
+};
+
+CardSearch.defaultProps = {
+  cards: [],
+  onSelect: () => true
+};
 
 export default CardSearch;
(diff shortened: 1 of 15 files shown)


next/components/card-search.js
```diff
@@ -0,0 +1,129 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import Autosuggest from 'react-autosuggest';
+import CardList from './card-list.js';
+
+const getSuggestions = (value, cards) => {
```
lily-pytel: I am a fan of putting functions like these in lib and unit testing - though I believe you are more in favor of integration, right?

next/components/card-search.js
```diff
@@ -0,0 +1,129 @@
...
+      id: PropTypes.number,
+      name: PropTypes.string
+    })
+  ),
+  onSelect: PropTypes.func
+};
+
+CardSearch.defaultProps = {
+  cards: [],
+  onSelect: () => true
```
jtrussell: I'm just curious, is this your favorite noop? I've never seen `() => true` before.

next/components/card-search.js
alexmiddeleer: No, that's a very good point. I like unit tests, I just haven't for whatever reason bothered with them at all yet.

next/components/deck-search-form.js
```diff
@@ -1,21 +1,32 @@
 import React from 'react';
+import { Query } from 'react-apollo';
+import ErrorMessage from './error-message';
 import PropTypes from 'prop-types';
 import { handleInputChange } from '../lib/form-utils';
+import CardSearch from './card-search';
+import { cardsQuery } from './all-cards.js';
```
lily-pytel: If you merge with `master`, you can replace this with the query from here: `https://github.com/mythgard-hub/mythgard-hub/blob/master/next/lib/queries/all-cards-query.js` - also you were the one who asked to give this same query a more specific name in your comment on my PR 😆

next/components/card-search.js
alexmiddeleer: I probably write a different one every time without thinking about it. I should really use a noop snippet.

next/components/card-search.js
```diff
@@ -0,0 +1,129 @@
...
+          suggestions={suggestions}
+          onSuggestionsFetchRequested={this.onSuggestionsFetchRequested}
+          onSuggestionsClearRequested={this.onSuggestionsClearRequested}
+          onSuggestionSelected={this.onSuggestionSelected}
+          getSuggestionValue={getSuggestionValue}
+          renderSuggestion={renderSuggestion}
+          inputProps={inputProps}
+          highlightFirstSuggestion={true}
+        />
+        <CardList
```
lily-pytel: Ah, I see that if an X is added, we won't be able to use this component as it is. I would vote that it's worth it since I made the mistake of clicking the card a couple of times. But definitely can figure this out later when there are more styles

next/components/deck-search-form.js
alexmiddeleer: haha yeah that's what CR is for right? calling out hypocrisy.

next/components/card-search.js
alexmiddeleer: We can style it with an x, potentially, using some fancy-ass css (content:before or somesuch). This component may not be enough, long-term.

alexmiddeleer: I will make it less horrible in this PR though since we're getting close to showing it to peeps.

jtrussell: A plug for accessibility - we should use a button ;). Also pseudo elements can be a little quirky when it comes to mouse interaction and focus.

alexmiddeleer:
@lily-pytel and @jtrussell Might need a little final peek before I merge to make sure the changes are up to snuff.

next/jest.config.js
```diff
@@ -2,7 +2,7 @@ module.exports = {
   moduleFileExtensions: ['js'],
   testMatch: ['**/*.(test|spec).js'],
   coveragePathIgnorePatterns: ['/node_modules/', 'enzyme.js'],
-  setupTestFrameworkScriptFile: '<rootDir>/enzyme.js',
+  setupFilesAfterEnv: ['<rootDir>/enzyme.js'],
```
jtrussell: Good catch!

alexmiddeleer: Just being a code monkey fixing a warning by taking the action it told me to do 😛

PR #26, opened 2019-08-11
Theme proof of concept (background + border)

alexmiddeleer:


next/components/layout.js
```diff
@@ -1,29 +1,52 @@
...
+          a {
+            color: ${theme.fontColor};
+          }
+        `}</style>
+      </div>
+    );
+  }
+}
+
+Layout.contextType = ThemeContext;
```
jtrussell: Neat! I haven't used `Component.contextType = ...`  since the context API updated. It looks like this is pretty similar to wrapping the component in `<ThemeContext.Consumer>context => (...)</ThemeContext.Consumer>` with the exception that you only get to access one context. Pretty cool.

alexmiddeleer: Yeah. Mostly just following react docs for that one. I think you're basically spot on.

PR #27, opened 2019-08-12
Kubernetify

jtrussell:
These changes enable the k8s deployment. Some things to note:

- There's a new `.env` variable, `PGSSL`, you can update your actual `.env.` file with the template value.
- There's a new `.env` variable, `API_HOST`, you can update your actual `.env.` file with the template value.
- The docker-compose file is no longer mounting `next/.next` as a volume. Was anyone making use of this? As a general rule we should avoid mounting resources that are purely derived from things already in the  container.
- Dockerfiles (for the `next` and `express` services in particular) should prep images for production rather than dev. Use docker-compose to make adjustments fo dev (k8s uses the Dockerfiles but not docker-compose files). If this gets to be annoying, we can have separate Dockerfiles for dev and prod envs. Right now, the only way this really impacts us is that the next service includes `npm run build` in its Dockerfile which amounts to just waiting for funsies if you're going to start the server in dev mode.
- Always use env variables to reference things like the server host names or ports.

There are separate k8s configs that are not included in this PR. I've still got to carve out secrets in those configs into [secrets](https://kubernetes.io/docs/concepts/configuration/secret/)... and I'm not even 100% sure those configs belong in this repo. In any event I'd like to get (and keep) this repo in k8s deployment shape.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -36,9 +36,7 @@ services:
     env_file:
       - .env
     volumes:
-      - type: bind
-        source: ./next
-        target: /app
       - /app/node_modules
     ports:
       - "3001:3001"
+    command: ["npm", "run", "dev"]
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -36,6 +36,9 @@ services:
     env_file:
       - .env
     volumes:
+      - type: bind
+        source: ./next
+        target: /app
       - /app/node_modules
     ports:
       - "3001:3001"
--- a/next/.dockerignore
+++ b/next/.dockerignore
@@ -1,5 +1,8 @@
-node_modules/*
-.eslintrc
-.prettierrc.yml
+.babelrc
+.dockerignore
+.eslintrc.json
 .gitignore
+.prettierrc.yml
+.next
 build.sh
+node_modules/*
--- a/next/Dockerfile
+++ b/next/Dockerfile
@@ -1,16 +1,8 @@
 FROM node:lts-alpine
 
 WORKDIR /app
+COPY . /app
 
-COPY package* ./
+RUN npm i --production --no-optional && npm run build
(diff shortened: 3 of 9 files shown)


next/.dockerignore
```diff
@@ -1,5 +1,8 @@
-node_modules/*
-.eslintrc
-.prettierrc.yml
+.babelrc
+.dockerignore
```
jtrussell: Huh, I bet this is redundant.

next/Dockerfile
```diff
@@ -1,16 +1,8 @@
 FROM node:lts-alpine
 
 WORKDIR /app
+COPY . /app
```
jtrussell: Get all the things by default and rely on `.dockerignore` to blacklist what we don't want.

next/Dockerfile
```diff
@@ -1,16 +1,8 @@
...
-
-COPY server.js .
-COPY pages .
-COPY components .
-COPY lib .
-COPY server-routes .
-COPY next.config.js .
-
-CMD ["npm", "run", "dev"]
+CMD ["npm", "start"]
```
jtrussell: Assume a prod env in `Dockerfile`s.

docker-compose.yml
```diff
@@ -36,9 +36,7 @@ services:
     env_file:
       - .env
     volumes:
-      - type: bind
-        source: ./next
-        target: /app
       - /app/node_modules
     ports:
       - "3001:3001"
+    command: ["npm", "run", "dev"]
```
jtrussell: Assume a dev env in `docker-compose.yml` files.

jtrussell:
I appreciate the vote of confidence :D.

As long as none of these changes look terrible, I'll be a happy camper.

PR #28, opened 2019-08-12
Add Travis CI for automated test running

evsheffield:
Here's a first pass at a configuration for Travis CI that will automatically run unit and e2e tests. Once all the DO stuff is sorted I'd love to chat with Justin about how I can set this up to handle our deploys as well. :)

We could probably get fancier about node or docker settings to optimize run times, but even this base configuration seems to do a decent job using caches to keep the npm install times down.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ node_js:
 
 sudo: required
 
-# Cypress seems to think this is needed :shrug:
+# Supposedly this is needed for Cypress to work in Ubuntu 16
 # https://github.com/cypress-io/cypress-example-kitchensink/blob/master/basic/.travis.yml
 addons:
   apt:
@@ -44,7 +44,7 @@ script:
   - cd next
   - npm run test
   # Run e2e tests
-  - cd e2e
+  - cd ../e2e
   - npx cypress verify
   - npx cypress run -- --record
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,6 @@
 language: node_js
+node_js:
+  - "10"
 
 sudo: required
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,12 @@ before_install:
   - cp .env.template .env
 
 install:
+  - cd next
+  - npm install
+  - cd ..
+  - cd e2e
+  - npm install
+  - cd ..
   - docker-compose up --build -d
 
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ services:
 
 before_install:
   # Upgrade docker-compose to the desired version
-  - sudo rm /usr/local/bin/docker-compoose
+  - sudo rm /usr/local/bin/docker-compose
   - curl -L https://github.com/docker/compose/releases/download/${DOCKER_COMPOSE_VERSION}/docker-compose-`uname -s`-`uname -m` > docker-compose
   - chmod +x docker-compose
   - sudo mv docker-compose /usr/local/bin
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -0,0 +1,52 @@
...
+  - sudo mv docker-compose /usr/local/bin
+  # Upgrade docker
+  - sudo apt-get update
+  - sudo apt-get -y -o Dpkg::Options::="--force-confnew" install docker-ce
+  # Put the .env file in place
+  - cp .env.template .env
+
+install:
+  - cd next
+  - npm install
```
jtrussell: These `npm install`s should be necessary.

.travis.yml
```diff
@@ -0,0 +1,52 @@
...
+  # Run unit tests
+  - cd next
+  - npm run test
+  # Run e2e tests
+  - cd ../e2e
+  - npx cypress verify
+  - npx cypress run -- --record
+
+post_script:
+  - docker-compose stop
```
jtrussell: Hmm... I might be misunderstanding how Travis will handle subsequent runs. I think it gets a fresh checkout + install each time an automated run is triggered. If not - and if we intent to reuse the build workspace for subsequent runs, then I might argue for `docker-compose down -v` over `docker-compose stop` so we're less likely to get gummed up by lingering container state. But, if the post_script here is just cleanup and we're not reusing the workspace then pay this no mind :D.

.travis.yml
evsheffield: They seemed to be in my test builds. 🤔 I figured it was because the dockerfiles only install non-dev dependencies and I didn't want to mess with those. But if you think it's alright to do so I can try to adjust them so everything we need is in place.

.travis.yml
```diff
@@ -0,0 +1,52 @@
+language: node_js
+node_js:
+  - "10"
```
jtrussell: This is trivial, but we could use `"lts/*"` instead of `"10"` to be consistent with the version currently listed in our Dockerfiles. I'm not sure we get many points for syncing up those versions... and to be honest we should probably lock down the Dockerfile node versions to a fixed number anyway 🤷‍♂.

evsheffield:
From what I've seen so far it doesn't seem to be re-using the workspace, so
even this may be unnecessary. I'll definitely keep that in mind in case I
come across any weirdness. TBH I mostly included it because I saw it in
various docker compose example Travis files and figured we may want to keep
it as a cleanup step for when we add deploy steps following it. ;)

.travis.yml
jtrussell: Using `npx` *should* also get you any deps that cypress needs to run. If things are failing without an `npm install` in the host that might point to other issues we'd want to shore up. FWIW when I run cypress there isn't even a node_modules folder in the `e2e` directory.

evsheffield:
Once i get the builds running for the main repo I'll add a Travis ci build
badge to the readme and any other info that might be needed for us all to
check out the builds for the project. A lot of stuff should be automatic,
such as Travis reporting whether PR branches pass the build. It's all
hosted so we don't need to set anything up and it's completely free for
open source projects.

I think to activate builds for the repo once this is merged I may need to
be a member of the mythgard-hub org on GitHub. Mark accepted my request to
allow Travis to connect to the organization but I still can't activate it
because I don't have permissions to do so. So if you have the ability to
add me that would be great, otherwise I'll ask Mark if he can get me
squared away. :)

.travis.yml
evsheffield: You are indeed correct about it being unnecessary for the e2e tests. Unfortunately as currently configured it does seem necessary for the jest unit tests, though I'm sure we could work that out later.

lily-pytel: Sorry about that :-(

PR #29, opened 2019-08-14
Fonts and text colors

alexmiddeleer:
This feels like a lot of css in the components. Is it recommended to stick these in different files or is that just normal in react world?

### DIFF
--- a/next/components/header.js
+++ b/next/components/header.js
@@ -1,6 +1,8 @@
 import Link from 'next/link';
 import { ThemeContext } from './theme-context';
 
+const cdn = `${process.env.MG_CDN}mgh/`;
+
 const linkStyle = {
   marginRight: 15
 };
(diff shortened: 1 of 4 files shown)


next/components/header.js
```diff
@@ -1,56 +1,110 @@
 import Link from 'next/link';
+import { ThemeContext } from './theme-context';
+
+const cdn = `${process.env.MG_CDN}mgh/`;
```
jtrussell: I don't think we have an entry in the `.env.template` file for this yet. And (right or wrong) as a point of consistency our host environment variables haven't included a trailing slash.

next/components/header.js
```diff
@@ -1,56 +1,110 @@
...
+          }
+          .header .home .crownIcon {
+            height: 50px;
+            vertical-align: top;
+          }
+          .header a {
+            text-decoration: none;
+          }
+          .header a:before {
+            content: '▶';
```
jtrussell: Could I convince you to use the unicode escape sequence here instead of the literal value? I feel like they randomly show up as question marks or "unknown char" boxes for me.

next/components/header.js
```diff
@@ -1,56 +1,110 @@
...
+    let theme = this.context;
+    return (
+      <div className="header">
+        <style jsx>{`
+          .header {
+            background: ${theme.background};
+            border-bottom: ${theme.border};
+          }
+        `}</style>
+        <style jsx>{`
```
jtrussell: Just curious, is there a reason for separate `style` blocks here?

alexmiddeleer: It is apparently better for performance to have the dynamic pieces in a different block. 
Citation: (https://github.com/zeit/styled-jsx#dynamic-styles)

next/components/header.js
alexmiddeleer: It's in next.config.js currently. Is that cool, or should I move it elsewhere?

PR #30, opened 2019-08-15
Fix a11y-unfriendly buttons

alexmiddeleer:


next/components/card-list-item.js
```diff
@@ -4,11 +4,20 @@ import PropTypes from 'prop-types';
...
+        <span data-cy="cardListCard" onClick={e => onClick(e, card)}>
+          {card.id}. name: {card.name}
+        </span>
+      )}
+      {!onClick && (
+        <Link href={`/card?id=${card.id}`}>
+          <a data-cy="cardListCard">
+            {card.id}. name: {card.name}
+          </a>
+        </Link>
```
alexmiddeleer: This is not very DRY but the contents are basically a glorified placeholder anyway.

lily-pytel: So if it's clickable, should it not be a button instead of a span? Although I imagine in the end it'll be a card image or something.

PR #31, opened 2019-08-16
Og headers

alexmiddeleer:


next/next.config.js
```diff
@@ -2,7 +2,7 @@ module.exports = {
   env: {
     GA_TRACKING_ID: 'UA-142178156-1',
     USE_GOOGLE_ANALYTICS: process.env.NODE_ENV === 'production',
-    MG_CDN: 'https://mythgardhub.s3-us-west-2.amazonaws.com/'
+    MG_CDN: 'https://mythgardhub.s3-us-west-2.amazonaws.com'
```
jtrussell: Let's add this to this `.env` file instead. We should probably just get rid of the `env` portion of the next config.

next/pages/_app.js
```diff
@@ -22,6 +24,20 @@ class MyApp extends App {
...
             />
+
+            <meta
+              property="og:title"
+              content="Mythgard Hub | Community Hub for Mythgard Card Game"
+            />
+            <meta property="og:site_name" content="Mythgard Hub" />
+            <meta
+              property="og:description"
+              content="A fan-built site for Mythgard including Decklists, Card Database and more."
```
evsheffield: Little nitpick but maybe "a Card Database" reads better?

next/next.config.js
alexmiddeleer: I didn't fully do this since I was concerned it wouldn't work properly in the browser. However I did move all the things to .env and updated the template. We can iterate if there are further improvements available.

next/components/header.js
```diff
@@ -63,10 +63,14 @@ class Header extends React.Component {
             >
               <img
                 className="crownIcon"
-                src={`${cdn}MGH_CrownLogo.png`}
+                src={`${cdn}/mgh/MGH_CrownLogo.png`}
+                alt=""
```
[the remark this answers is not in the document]
alexmiddeleer: done

next/pages/_app.js
alexmiddeleer: fixed and removed strange capitals

PR #32, opened 2019-08-16
Add k8s deploy scripts

jtrussell:
I'm sure there are things in here that are not totally, 100%, all the way, iron clad, best practices.

This is still a manual deploy process, and requires yet another file that is not tracked by version control. Basically the same thing as `.env` except with production values.

Updated the README to include steps to deploy though there is some other configuration you'll need to do as well (mostly to set up the `kubectl` utility). I'll share the prod secrets file with you all too.

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -8,7 +8,7 @@ GOOGLE_CALLBACK_URL=http://localhost:3001/auth/google/callback
 GOOGLE_AUTH_SCOPE=https://www.googleapis.com/auth/userinfo.email
 
 # Google Analytics
-GA_TRACKING_ID=UA-142178156-1
+GA_TRACKING_ID=XX-YYYYYYYYY-1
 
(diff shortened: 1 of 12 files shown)


jtrussell:
Flagging https://github.com/mythgard-hub/mythgard-hub/pull/32/commits/456e4befe65633d4e1f38b3a95d12dfae20aaf29 as a commit of interest. The rational here is that we want to have access to certain environment variables when `next build` runs. I'm not sure there's a great way of doing this without smashing those variables into yet another spot (`ENV` directives int the Dockerfile)... which would mean they'd then live in the Dockerfile, `.env` and k8s secrets. Building in the deployed-to environment means you should have access to the correct env vars at build time.

Definitely not ideal. It might make sense to move all non-sensitive things to Dockerfiles (all the things Next cares about) and use secrets just for secrets.

PR #33, opened 2019-08-17
Import cards

alexmiddeleer:


lily-pytel:
We should change all the janky e2e tests to work with real cards - yay!

alexmiddeleer:
@lily-pytel Yeah it's not part of the Docker build. I didn't want the card data to be freely available.

PR #34, opened 2019-08-17
No need to record videos with travis cypress tests

alexmiddeleer:


alexmiddeleer:
well, that didn't work.

PR #35, opened 2019-08-18
Deck builder styles

lily-pytel:
- Added global styles for buttons and inputs of type submit to match wireframe
- Moved mutations and mutation shells to their own files. These will be refactored a bit once I learn to use `useMutation`
- Created a component for the deck name. It just has an input for now, but it'll be a bit nicer (showing an editable title)
- Refactored `DeckBuilderPage` to use hooks instead of state, moved the parts around to match the wireframe more and other general refactoring

### DIFF
--- a/next/components/edit-deck-name.js
+++ b/next/components/edit-deck-name.js
@@ -1,0 +1,31 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export default function EditDeckName(props) {
+  const { deckName, onChange } = props;
+
+  return (
+    <div>
+      <label>
+        Deck Name:{' '}
+        <input
+          data-cy="deckTitle"
+          type="text"
+          name="deckName"
+          onChange={onChange}
+          value={deckName}
+        />
+      </label>
+    </div>
+  );
+}
+
+EditDeckName.defaultPros = {
+  deckName: '[new deck]',
+  onChange: () => {}
+};
+
+EditDeckName.propTypes = {
+  deckName: PropTypes.string,
+  onChange: PropTypes.func
+};
--- a/next/components/edit-deck-name.js
+++ b/next/components/edit-deck-name.js
@@ -21,8 +21,7 @@ export default function EditDeckName(props) {
 }
 
 EditDeckName.defaultProps = {
-  deckName: '[new deck]',
-  onChange: () => {}
+  deckName: '[new deck]'
 };
 
 EditDeckName.propTypes = {
--- a/next/components/edit-deck-name.js
+++ b/next/components/edit-deck-name.js
@@ -20,7 +20,7 @@ export default function EditDeckName(props) {
   );
 }
 
-EditDeckName.defaultPros = {
+EditDeckName.defaultProps = {
   deckName: '[new deck]',
   onChange: () => {}
 };
--- a/next/components/layout.js
+++ b/next/components/layout.js
@@ -49,6 +49,18 @@ class Layout extends React.Component {
           a {
             color: ${theme.fontColor};
           }
+
+          button,
+          input[type='submit'] {
+            background-color: ${theme.buttonBackground};
+            border: ${theme.sectionBorder};
+            color: ${theme.fontColor};
+            font-family: ${theme.fontFamily};
+            font-size: 20px;
+            padding: 5px 10px;
+            width: 100%;
+            border-radius: 10px;
+          }
         `}</style>
       </div>
     );
--- a/next/components/theme-context.js
+++ b/next/components/theme-context.js
@@ -5,8 +5,8 @@ export const themes = {
   dark: {
     background: mgColors.blackPearl,
     fontColor: mgColors.white,
-    buttonBackground: mgColors.buttonBackground,
-    sectionBorder: `2px solid ${mgColors.sectionBorderColor}`,
+    buttonBackground: mgColors.mediumGray,
+    sectionBorder: `2px solid ${mgColors.lightBlue}`,
     border: `2px solid ${mgColors.orange}`,
     fontFamily: mgFonts.exo2
   }
--- a/next/components/theme-context.js
+++ b/next/components/theme-context.js
@@ -5,6 +5,8 @@ export const themes = {
   dark: {
     background: mgColors.blackPearl,
     fontColor: mgColors.white,
+    buttonBackground: mgColors.buttonBackground,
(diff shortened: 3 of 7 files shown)


next/components/edit-deck-name.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
```
lily-pytel: This component will be more than just this input but either way too many lines to not be a component

next/components/layout.js
```diff
@@ -49,6 +49,18 @@ class Layout extends React.Component {
           a {
             color: ${theme.fontColor};
           }
+
```
lily-pytel: Makes buttons like this: 
image

next/lib/mutations/add-card-to-deck.js
```diff
@@ -0,0 +1,35 @@
+import gql from 'graphql-tag';
```
lily-pytel: I usually only put the gql itself in one of these files, but in this case it seemed like why not put the shell as well?

next/pages/deck-builder.js
```diff
@@ -1,66 +1,18 @@
-import React from 'react';
```
lily-pytel: The diff viewer is a mess when you convert a class component into a function component... sorry! I recommend just looking at the file. I'm sorry I said mean things about hooks. I still like redux for redux things but hooks are definitely a dope way to get rid of state and class components, AND APOLLO HOOKS!!! To be shown in the next PR

next/components/edit-deck-name.js
```diff
@@ -0,0 +1,31 @@
...
+          name="deckName"
+          onChange={onChange}
+          value={deckName}
+        />
+      </label>
+    </div>
+  );
+}
+
+EditDeckName.defaultPros = {
```
alexmiddeleer: typo props

next/components/edit-deck-name.js
```diff
@@ -0,0 +1,31 @@
...
+  deckName: '[new deck]',
+  onChange: () => {}
```
alexmiddeleer: we may prefer an error if onchange is forgotten, so could nix defaultprop here

next/lib/theme.js
```diff
@@ -3,7 +3,9 @@ export const mgColors = {
   orange: '#F1810B',
   blue: '#4EB8E7',
   white: '#FFF',
-  lightGray: '#D1D1D1'
+  lightGray: '#D1D1D1',
+  buttonBackground: '#333333',
```
alexmiddeleer: mgColors should only contain names like "orange". It's the "no-abstraction" level.

next/lib/theme.js
```diff
@@ -3,7 +3,9 @@ export const mgColors = {
...
+  sectionBorderColor: '#458A9E'
```
alexmiddeleer: ditto. Can just be something like "gray_1" if necessary.

next/components/edit-deck-name.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export default function EditDeckName(props) {
+  const { deckName, onChange } = props;
+
+  return (
+    <div>
+      <label>
+        Deck Name:{' '}
```
alexmiddeleer: the cool kid's way of doing `&nbsp;`? 😆

next/components/theme-context.js
```diff
@@ -5,6 +5,8 @@ export const themes = {
   dark: {
     background: mgColors.blackPearl,
     fontColor: mgColors.white,
+    buttonBackground: mgColors.buttonBackground,
+    sectionBorder: `2px solid ${mgColors.sectionBorderColor}`,
```
alexmiddeleer: see comments on theme.js

next/lib/mutations/add-deck.js
```diff
@@ -0,0 +1,21 @@
...
+    createDeck(input: { deck: { name: $name } }) {
+      deck {
+        name
+        id
+      }
+    }
+  }
+`;
+
+const createDeckShell = (apolloClient, deckName) => {
```
alexmiddeleer: The name `createDeckShell` is confusing to me. Can we just call it `createNewEmptyDeck`?

next/pages/deck-builder.js
```diff
@@ -76,162 +28,127 @@ const saveDeck = (apolloClient, deckInProgress) => {
...
     });
-  }
+  };
 
-  onCollectionClick(e, card) {
-    const { deckInProgress } = this.state;
+  const onCollectionClick = (e, card) => {
     e && e.preventDefault();
 
     const nextMainDeck = addCardToDeck(deckInProgress.mainDeck, {
```
alexmiddeleer: Sorry for all the nitpicks but I like `newMainDeck` instead of `nextMainDeck`. Next, to me, implies "next in the list of things". As I get older I get super anal about variable names. I know you didn't even make this name in this PR...

next/components/edit-deck-name.js
lily-pytel: It's the next level of props - the pros league

next/components/edit-deck-name.js
lily-pytel: Sounds good

next/lib/theme.js
lily-pytel: I am not a fan but I shouldn't have tried to defy the clear convention - my bad!!!

next/components/edit-deck-name.js
lily-pytel: I thought you wrote that! It was there before 😆 . Anyway, probably going to get rid of it tomorrow altogether and make it an editable title

next/lib/mutations/add-deck.js
lily-pytel: I also feel like you created it 😆 But yes, we can

next/pages/deck-builder.js
lily-pytel: I did make it in a previous PR though so no worries. Will change

next/lib/theme.js
alexmiddeleer: I will try to make you a fan at some point - it's worth it, I promise.

next/lib/mutations/add-deck.js
alexmiddeleer: Don't critique the coder, critique the code 😆

jtrussell:
Long the the hook.

PR #36, opened 2019-08-18
Import path power validation

lily-pytel:
- `import-deck` converted to working with apollo hooks
- path and power validation added. They are excluded if they don't exist in our database

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,7 +92,7 @@ Please use prettier and eslint. Configs are located in next, express, and e2e, b
 
 ## Deploying
 
-Ensure that the `docker` and `kubectl` commands are installed and in you path.  Docker will need to be configured to push to the `gsdf` organization in the public registry and `kubectl` must be pointed at your target cluster.
+Ensure that the `docker` and `kubectl` commands are installed and in you path. Docker will need to be configured to push to the `gsdf` organization in the public registry and `kubectl` must be pointed at your target cluster.
 
 First, create a release. E.g.:
 
@@ -115,7 +115,3 @@ Note that in order to run this script you will have to create the file `kube-con
 ## Importing Data
 
 See comments for import-\* scripts in /scripts
-
-## Misc Notes
-
-- To remove all containers: `docker container rm $(docker container ps -aq)`
--- a/README.md
+++ b/README.md
@@ -90,7 +90,6 @@ To run all tests:
 
 Please use prettier and eslint. Configs are located in next, express, and e2e, but you will need to `npm install` the necessary eslint plugins.
 
-
 ## Deploying
 
 Ensure that the `docker` and `kubectl` commands are installed and in you path.  Docker will need to be configured to push to the `gsdf` organization in the public registry and `kubectl` must be pointed at your target cluster.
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -115,3 +115,7 @@ Note that in order to run this script you will have to create the file `kube-con
 ## Importing Data
 
 See comments for import-\* scripts in /scripts
+
+## Misc Notes
+
+- To remove all containers: `docker container rm $(docker container ps -aq)`
```
alexmiddeleer: I'd probably remove this as it's not specific to mythgard.  We could create another markdown file called `docker-tips.md`, I'd be okay/happy with that. But then we need to maintain it.

next/lib/queries/paths-powers-query.js
```diff
@@ -0,0 +1,20 @@
+import gql from 'graphql-tag';
+
+const pathsAndPowersQuery = gql`
```
alexmiddeleer: You may prefer to make these separate queries since it's more likely to be cachable. We use query batching so it's 1 network request either way. Though, on second thought, that's probably not the graphql way. I'll leave this comment for discussion but you can ignore it if you want. I'm not sure either way.

next/pages/_app.js
```diff
@@ -17,29 +18,31 @@ class MyApp extends App {
...
       <Container>
         <ApolloProvider client={apolloClient}>
-          <Head>
-            <meta charSet="utf-8" />
-            <meta
-              name="viewport"
-              key="viewport"
-              content="width=device-width, initial-scale=1"
-            />
+          <HooksApolloProvider client={apolloClient}>
```
alexmiddeleer: Haha hooks were supposed to remove excessive nesting, but I guess you need just one more first 😉

next/lib/queries/paths-powers-query.js
lily-pytel: I do find separate queries more logical and when possible I do prefer them - like in this case, where our queries are batched. I recognize it's probably not ideal and I should learn to use fragments but I haven't yet

next/pages/_app.js
lily-pytel: Just one more to save a million more!

lily-pytel: Plus the other one will be removed when (not if!) we get rid of all the other ways of doing it

README.md
lily-pytel: Fair enough, I put it in my own notes

next/lib/queries/paths-powers-query.js
lily-pytel: Oh, I totally missed the point of the question. Yeah, I might separate them in the next PR

next/pages/_app.js
```diff
@@ -3,6 +3,7 @@ import Head from 'next/head';
 import Router from 'next/router';
 import withApolloClient from '../components/with-apollo-client';
 import { ApolloProvider } from 'react-apollo';
+import { ApolloProvider as HooksApolloProvider } from 'react-apollo-hooks';
```
jtrussell: Will we need both providers in the long run?

lily-pytel: Hell no!
(I'm hoping. I haven't gotten to the point of replacing all the old components with hooks yet)

PR #37, opened 2019-08-18
Search decks by color

alexmiddeleer:


alexmiddeleer:
I need to work on this a bit more. The react-apollo upgrade really ruined my night 😢

next/components/create-comment.js
```diff
@@ -43,15 +44,14 @@ class CreateComment extends React.Component {
...
+    const query = this.props.deckCommentsQuery;
 
-    previousCommentNodes.push(deckComment);
-    cache.writeQuery({
-      query: this.props.deckCommentsQuery,
-      data: deckQueryObj
-    });
+    const data = cache.readQuery({ query, variables });
+
+    const newData = _.cloneDeep(data);
```
alexmiddeleer: I know this is gross. I pulled my hair out trying to not do it this way, but it seems to be necessary. See https://github.com/apollographql/apollo-client/issues/3909 if curious.

alexmiddeleer:
test failure seems to be a false positive.

next/components/create-comment.js
lily-pytel: Ugh, I'm having flashback to a few weeks ago. This wouldn't be fixed with using the new hooks, would it? Probably not. I've actually had a new behavior with the cache at work when I used the hooks, so it's a bit unpredictable

lily-pytel:
> I need to work on this a bit more. The react-apollo upgrade really ruined my night 😢

@alexmiddeleer I'm so sorry!!! What changed?

alexmiddeleer:
> > I need to work on this a bit more. The react-apollo upgrade really ruined my night 😢
> 
> @alexmiddeleer I'm so sorry!!! What changed?

It wasn't you, it was me. I updated react-apollo. See commit for fix stuck loading messages.

lily-pytel:
> > > I need to work on this a bit more. The react-apollo upgrade really ruined my night 😢
> > 
> > 
> > @alexmiddeleer I'm so sorry!!! What changed?
> 
> It wasn't you, it was me. I updated react-apollo. See commit for fix stuck loading messages.

@alexmiddeleer I thought it was my fault at first and then I saw that you updated in this PR. My three exclamations were for the pain of updating anything

alexmiddeleer:
> @alexmiddeleer I thought it was my fault at first and then I saw that you updated in this PR. My three exclamations were for the pain of updating anything. It was you who taught me the motto "never update"
Well when the current version has a horrible bug it kinda evens it out. The new bug was bad though... killed a few hours and made me a lot more wary of react-apollo.

lily-pytel:
@alexmiddeleer do you want eyes on this before resolving the little conflict with the node modules? I would pull master into this branch just in case our changes clashed somehow

next/components/faction-filter.js
```diff
@@ -1,12 +1,14 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import { ThemeContext } from './theme-context';
```
lily-pytel: So this can't be used in function components? Or do they need to be wrapped by some provider?

next/components/create-comment.js
```diff
@@ -4,6 +4,7 @@ import gql from 'graphql-tag';
 import { Mutation } from 'react-apollo';
 import ErrorMessage from './error-message';
 import { handleInputChange } from '../lib/form-utils';
+import _ from 'lodash';
```
jtrussell: Can we do a cherry pick import? Something like `import deepClone from 'lodash/foo/deepClone'`? It should result in a smaller bundle getting shipped to the browser.

jtrussell: Also, did lodash make it into the package.json as a dep?

next/components/faction-filter.js
jtrussell: With a consumer you could use an FC. A matter a style I guess ¯\_(ツ)_/¯.

next/components/create-comment.js
alexmiddeleer: I actually tried this and the bundle size was no different. I was surprised actually. I even installed lodash-es and next-transpile-modules. I think tree-shaking may already be happening via the magic of webpack. I could have also just borked it.

jtrussell: @alexmiddeleer Webpack's tree shaking will only key in on what's explicitly imported and exported. Importing all of lodash gets you about an extra .4MB sent to the browser in dev mode. Haven't checked prod. See the image below.

Screen Shot 2019-08-20 at 8 47 55 PM

The first two lines are with `import cloneDeep from 'lodash/cloneDeep';` The last line is with `import _ from 'lodash';`.

Also, we really should add `lodash` as a dep to the next app.

alexmiddeleer:
Alright, will get both the cherry pick and dep in tomorrow.

alexmiddeleer:
@jtrussell just a note - I haven't forgotten about this. But comments are ded now.

jtrussell:
Not worried about it - anything that works gets a green light at this stage ;).

PR #44, opened 2019-08-24
Search cards by text

alexmiddeleer:


next/lib/card-queries.js
```diff
@@ -0,0 +1,68 @@
...
+  return `
+    or: [{
+      name: {
+        includesInsensitive: $searchText
+      }
+    },{
+      rules: {
+        includesInsensitive: $searchText
+      }
+		}]
```
jtrussell: Tabs snuck in?

next/pages/cards.js
```diff
@@ -1,14 +1,45 @@
...
+      >
+        <h1>Cards</h1>
+        <CardSearchForm onSubmit={this.handleSearchSubmit.bind(this)} />
+        <SomeCards filters={this.state.searchQuery} />
+      </Layout>
+    );
+  }
+}
+
+CardsPage.propTypes = {
```
jtrussell: kill dis?

next/lib/card-queries.js
```diff
@@ -0,0 +1,68 @@
+import gql from 'graphql-tag';
+import { useQuery } from 'react-apollo-hooks';
+
+export const getFactionsFilter = factionNames => {
+  if (!factionNames.length) {
+    // null means ignore this filter.
+    // Comma allows chaining.
+    return 'cardFactions: null,';
```
jtrussell: My gut tells me a happier path would be to create a function on the postgres side that takes values from all inputs and handles nulls/empties there. Could be a fun opportunity to try something other than SQL as the function language.

Not worth a refactor if things are working, just something to keep in mind if building a dynamic graphql query starts to get unwieldy or we find that it's hard to protect this endpoint from bad actors.

alexmiddeleer: Probably true. I'm just trying to get the mvp done leveraging the techniques we have already been using. Definitely might need to rewrite but this wasn't that hard to just spin up. Happy to remove it later if it causes problems.

PR #45, opened 2019-08-24
Add search by text to deckbuilder

alexmiddeleer:
There's a todo to make this more DRY with the card-list page. I will probably sneak that into a later PR.


next/components/deck-builder-search-form.js
```diff
@@ -0,0 +1,34 @@
+import React, { useState } from 'react';
```
jtrussell: Looks like the `useState` hook didn't end up getting used.

PR #46, opened 2019-08-25
Add page banner component

jtrussell:
Note that this changes the `MG_CDN` env variable.

I made some very minor style changes to bring things closer to our mockups... mostly so it was easier to compare banner as I was adding styles for them.

I'm debating whether or not we should skip the background image for mobile browsers. Maybe @PapaCapricorn has opinions on this?

### DIFF
--- a/e2e/cypress/integration/deck-builder-page.js
+++ b/e2e/cypress/integration/deck-builder-page.js
@@ -13,7 +13,7 @@ describe('Deck builder page', () => {
     cy.visit('/deck-builder');
   });
   it('should have a happy path', function() {
-    cy.get('h1').should('be.visible');
+    cy.get('[data-cy="header"]').should('be.visible');
     cy.get('[data-cy="importDeckTitle"]').should('be.visible');
     cy.get('[data-cy="importDeckTextarea"]').should('be.visible');
     cy.get('[data-cy="importDeckButton"]').should('be.visible');
--- a/e2e/cypress/integration/deck-builder-page.js
+++ b/e2e/cypress/integration/deck-builder-page.js
@@ -12,7 +12,7 @@ describe('Deck builder page', () => {
     cy.visit('/deck-builder');
   });
   it('should have a happy path', function() {
-    cy.get('[data-cy="header"]').should('be.visible');
+    cy.get('h1').should('be.visible');
     cy.get('[data-cy="importDeckTitle"]').should('be.visible');
     cy.get('[data-cy="importDeckTextarea"]').should('be.visible');
     cy.get('[data-cy="importDeckButton"]').should('be.visible');
--- a/next/components/page-banner.js
+++ b/next/components/page-banner.js
@@ -1,0 +1,55 @@
+import { Component } from 'react';
+import { ThemeContext } from './theme-context';
+import PropTypes from 'prop-types';
+
+class PageBanner extends Component {
+  render() {
+    const { children, image } = this.props;
+    const theme = this.context;
+    return (
+      <div className="page-banner">
+        <style jsx>{`
+          .page-banner {
+            border-top: ${theme.border};
+            border-bottom: ${theme.border};
+            background: url(${image}) left no-repeat, ${theme.background};
+            height: 70px;
+            display: flex;
+            flex-direction: row-reverse;
+            align-items: center;
+          }
+          .page-banner h1 {
+            color: ${theme.fontColorSelected};
+            margin: 0;
+            padding-right: 0.5em;
+            text-transform: uppercase;
+          }
+        `}</style>
+        <h1>{children}</h1>
+      </div>
+    );
+  }
+}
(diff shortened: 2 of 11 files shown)


next/components/page-banner.js
```diff
@@ -0,0 +1,55 @@
+import { Component } from 'react';
+import { ThemeContext } from './theme-context';
+import PropTypes from 'prop-types';
+
+class PageBanner extends Component {
+  render() {
+    const { children, image } = this.props;
+    const theme = this.context;
```
lily-pytel: So we really are just doing class components whenever we need the theme?

lily-pytel:
JK it was all fixed with a computer restart. To run this PR do I really need to clone your version of this repo in another folder, go to the branch and then rebuild a whole new docker? Surely not.

next/components/page-banner.js
jtrussell: You can use a function component as well with the theme context consumer. I'll change this as an example. Function components are probably a better default if we're going with hooks everywhere.

jtrussell:
Yep, you can check out the PR like it's a branch. You could also add my fork as a remote endpoint if you wanted all the things.

```
git remote add jtrussell https://github.com/jtrussell/mythgardhub
git fetch jtrussell
git checkout jtrussell/branch
```

next/components/page-banner.js
jtrussell: I take that back, there's a `useContext` hook. We should just use that.

e2e/cypress/integration/deck-builder-page.js
```diff
@@ -12,7 +12,7 @@ describe('Deck builder page', () => {
     cy.visit('/deck-builder');
   });
   it('should have a happy path', function() {
-    cy.get('[data-cy="header"]').should('be.visible');
+    cy.get('h1').should('be.visible');
```
alexmiddeleer: don't need to worry about it but the data-cy thing is a best practice.

jtrussell: 👍

next/components/page-banner.js
```diff
@@ -0,0 +1,50 @@
+import { useContext } from 'react';
+import { ThemeContext } from './theme-context';
+import PropTypes from 'prop-types';
+
+function PageBanner({ children, image }) {
+  const theme = useContext(ThemeContext);
```
jtrussell: @lily-pytel Much better 👏.

lily-pytel: I love it! I've been highly suspicious of class components since using hooks.

lily-pytel: And this is so much better than the provider wrapper component OMG

jtrussell: WAY better. Class components are dtm.

lily-pytel: TIL about `dtm`

evsheffield: Kids these days and all their fancy acronyms.

jtrussell: 🚗 https://en.wikipedia.org/wiki/Deutsche_Tourenwagen_Masters 🚗

next/components/page-banner.js
PapaCapricorn: @jtrussell I don't feel strongly about skipping the background image for mobile users, but would be curious to know the rationale...Is it to maximize screen real estate? Minimize downloads? Both?

jtrussell: Both of those and also legibility.

Screen Shot 2019-08-25 at 11 00 29 PM

lily-pytel: Both and also just the fact that they will look like this on mobile
image

So you can't really even seen them, unless we have smaller versions that scale better for small screens.

PapaCapricorn: Oh, I see...I thought you were referring to the fixed background image on the borders. Will give this some more thought and respond...

PapaCapricorn: Thanks for that screen cap @jtrussell - it is really useful in clarifying what the issue is. (At first, I thought you were talking about the full-sized background at -1 z-axis...)

This may be my inexperience with responsive design for mobile showing, but can we not just scale down the entire banner (image and text) to a thinner band at the top? (See comp…)

The text overlapping the image is clearly a problem, but even if that were not the case - imo the banner takes up too much vertical space at full height anyway.

Comparison

jtrussell: That's a good idea, @PapaCapricorn - we can go that route. I'll update the style to scale the background image with the container size, we'll just want to keep an eye on the page weight. Generally, we want to avoid scaling images directly in the browser if we can avoid it.

As for text sizes, we may want to set values for those at different breakpoints, however we end up handling max-width on the page content. It might be worth us putting a story in the backlog to harden styles across the site - might take a couple sessions to hammer out for various widths/etc.

PR #47, opened 2019-08-25
Add account page and support for setting username

evsheffield:
There's definitely pieces relating to this that will still need work (e.g. checking if a username is already taken), but I've been sitting on it far too long with other stuff being blocked so I wanted to get it out there.

I've also made some updates to how the user is made available to the components. User is now available through a user context. You can still use the `withUser` HOC which is now essentially just a wrapper for the user context but the added benefit is that you can use it with any component instead of just pages. `_app.js` is now getting the user info ALL THE TIME which is helpful since we're gonna want it all about, but could probably be improved to prevent redundant requests. 

On a side note, if anyone has any advice for debugging SSR stuff I would love to hear it since I spent far too long smashing my head against `getInitialProps`. It doesn't seem like the server output includes console.logs so I was just feeling my way around the dark like a cave man.


alexmiddeleer:
I've had success just putting `debugger;` in - that seems to disable SSR.

jtrussell:
FWIW SSR console.log's should show up in the terminal.

next/pages/_app.js
```diff
@@ -13,8 +16,55 @@ if (USE_GOOGLE_ANALYTICS) {
 }
 
 class MyApp extends App {
+  static async getInitialProps({ ctx, Component }) {
+    const isSSR = !!ctx.req;
```
alexmiddeleer: I'm kind-of anti `!!` unless the value truly needs to be boolean for some reason.

next/pages/account.js
```diff
@@ -0,0 +1,84 @@
...
+
+  areSettingsValid = () => {
+    return !!this.state.username;
+  };
+
+  handleSubmit = (apolloClient) => {
+    if (!this.areSettingsValid()) return;
+    const { user, updateUser } = this.context;
+    const { username } = this.state;
+    updateUsername(apolloClient, user.id, username).then(({ data }) => {
```
alexmiddeleer: do we need any error handling code for this?

next/server-routes/auth.js
```diff
@@ -31,7 +44,9 @@ passport.use(
       const values = [id, email];
       client.query(text, values, (err, res) => {
         if (err) return cb(err);
-        cb(null, { email: res.rows[0].email });
+        cb(null, {
+          email: res.rows[0].email
```
alexmiddeleer: is there an issue if rows has a length of zero?

next/pages/account.js
```diff
@@ -0,0 +1,84 @@
...
+    });
+  };
+
+  render() {
+    const { user } = this.context;
+    const { username } = this.state;
+    return (
+      <Layout
+        title="Mythgard Hub | Account Settings"
+        desc="The Mythgard Hub privacy policy"
```
jtrussell: Looks like a copy+pasted page description might have snuck in there.

next/pages/account.js
```diff
@@ -0,0 +1,84 @@
...
+          <div>
+            <label>Email: </label>
+            <span>{user.email}</span>
+          </div>
+          <div>
+            <label>User Name: </label>
+            <input
+              data-cy="deckTitle"
+              type="text"
+              name="deckName"
```
jtrussell: Label: User Name; data-cy: deckTitle; name: deckName 😂.

next/server-routes/auth.js
jtrussell: It will (should?) never have length zero - the method is off-screen but it's job is to:
-  Find a thing if it exists
- Create a thing and return it if it doesn't exist
If either of those hit a snag `err` will be non-empty.

next/server-routes/auth.js
```diff
@@ -85,24 +100,24 @@ router.get('/logout', (req, res) => {
...
 });
 
-router.get('/user', (req, res) => {
+router.get('/user', async (req, res) => {
   const signedToken = req.cookies[process.env.JWT_COOKIE_NAME];
   let user;
   try {
     const token = jwt.verify(signedToken, process.env.JWT_SECRET);
-    user = token.data;
+    user = await getUserByEmail(token.data.email);
```
jtrussell: Alternatively, we could stuff additional detail into the token and avoid a round trip to the db. A tradeoff to consider when we're rich an famous with a ginormous DB of users.

next/pages/_app.js
jtrussell: I think this was my line moved from the old `with-user` code so I won't make Evan stand by it 😅. There wasn't a good reason other than habit to signal intent and save our plucky heroes from wondering if `undefined` meant "ok, no" or "oops, I forgot". Feel free to reclaim the bits.

evsheffield: Fair enough, this used to need to be a Boolean when it was being stored as a prop but it wasn't being used elsewhere so I can probably just do away with it.

next/pages/account.js
evsheffield: Oops hehe good catch

next/pages/account.js
evsheffield: Caught me red-handed

next/server-routes/auth.js
evsheffield: I could get down with that, though I guess we'd have to do a little more management of the token to make sure there's no stale data if the user updates their stuff mid session.

evsheffield:
> 
> 
> FWIW SSR console.log's should show up in the terminal (Docker side console).

So I should see that in the regular output that displays when I run `docker-compose up`? I could have sworn it wasn't showing logs that I thought would be there though entirely possible I had just broken things and they weren't even getting hit haha

next/pages/account.js
evsheffield: Right on. Do we have any standard way of raising errors to the user atm? Can certainly just put some console errors for now but some consistent user alerting would be sweet.

next/pages/account.js
lily-pytel: https://thumbs.gfycat.com/MerryOpenAmurstarfish-size_restricted.gif

next/server-routes/auth.js
jtrussell: This is one of those things where I feel like someone who has worked with JWTs a lot would have strong opinions but I'm just like `/shrug`. Like maybe the payload should be teeny and obscure just in cases.

next/server-routes/auth.js
lily-pytel: You are so trusting 😓

jtrussell:
> So I should see that in the regular output that displays when I run `docker-compose up`? I could have sworn it wasn't showing logs that I thought would be there though entirely possible I had just broken things and they weren't even getting hit haha

Yep. And if things are getting drowned out by noise from the other services you _can_ attach a new terminal to just the output of the next app. It's a lot of hoops to go through just for `console.log` output.

next/components/layout.js
```diff
@@ -68,6 +60,29 @@ class Layout extends React.Component {
...
+          .stack {
+            --space: 1.5rem;
+          }
+
+          .stack > * {
+            margin-top: 0;
+            margin-bottom: 0;
+          }
+
+          .stack > * + * {
```
lily-pytel: So much fancy CSS

next/server-routes/auth.js
jtrussell: I feel the suspicion, and tip my hat. If there's one thing I trust a SQL product to do, it's break loudly and with much fanfare. And especiallyl if CSV files are involved.

next/pages/account.js
alexmiddeleer: We have an ErrorMessage Component, but that's it.

next/components/layout.js
alexmiddeleer: the great lobotomized owl

alexmiddeleer:
I've been reading a bit about react security, I guess it doesn't escape code or other nastiness in props. I didn't have enough time to re-review this PR fully but it didn't seem to be doing anything dangerous. But we should keep in mind that the username is a cross site scripting risk. Could even see an argument for naming the username "dangerousUsername". Apparently that's a thing people do.

jtrussell:
> I've been reading a bit about react security, I guess it doesn't escape code or other nastiness in props. I didn't have enough time to re-review this PR fully but it didn't seem to be doing anything dangerous. But we should keep in mind that the username is a cross site scripting risk. Could even see an argument for naming the username "dangerousUsername". Apparently that's a thing people do.

Oh that's a great point. We should add a follow-up card to make sure usernames are safe to display. I'd be pretty happy with something as simple as white listing characters for now.

PR #48, opened 2019-08-25
Deckbuilder search cards by rarity

alexmiddeleer:


next/components/rarity-filter.js
```diff
@@ -0,0 +1,76 @@
...
+  const [isRare, setIsRare] = useState(false);
+  const [isMythic, setIsMythic] = useState(false);
+
+  const flags = [isCommon, isUncommon, isRare, isMythic];
+
+  if (previousRarities.join() !== flags.join()) {
+    onChange(flagsToEnums(flags));
+  }
+
+  previousRarities = [...flags];
```
jtrussell: Perhaps a good use case for `useEffect`:

```
useEffect(() => onChange(flagsToEnums(flags)), flags)
```

And we can get rid of tracking previous rarities manually.

alexmiddeleer: I am doing an overhaul in the next branch since it's virtually the same feature, so I will try this in that branch. Thanks for the tip!

PR #53, opened 2019-08-29
Cards page faction filter

alexmiddeleer:


lily-pytel:
@alexmiddeleer if you have no objections to making the card-search-form component functional: https://github.com/mythgard-hub/mythgard-hub/pull/55

alexmiddeleer:
> @alexmiddeleer if you have no objections to making the card-search-form component functional: #55

Thanks! I'm all about the hooks. And I agree that stateless is a little bit of  silly name so thanks for fixing it.

PR #54, opened 2019-08-29
Cards page type filter

alexmiddeleer:


lily-pytel:
Weird that the deck builder page test is failing on Travis when it seems just fine...?

PR #58, opened 2019-08-30
Delete decks

jtrussell:
I tried to mirror how the export feature worked in all ways that it was possible to mirror.

One thing to note - there's no BE validation for "should this user be able to delete this deck". I figured we could layer that on once https://trello.com/c/Ij1vcMTd/76 is complete.

Another thing I'd be interested to get input on, do you think we should redirect somewhere after deleting a deck? I have the user being left on the deck page for the deleted deck (you can't refresh the page without error). Toyed around with the idea of sending them to their profile page (i.e. *my* decks) or the all decks page... or somewhere else ?

### DIFF
--- a/next/components/deck.js
+++ b/next/components/deck.js
@@ -1,10 +1,14 @@
+import { useContext } from 'react';
+import { useQuery } from 'react-apollo-hooks';
 import PropTypes from 'prop-types';
 import gql from 'graphql-tag';
 import { Query } from 'react-apollo';
 import ErrorMessage from './error-message';
 import DeckCardList from './deck-card-list';
 import DeckExport from './deck-export';
+import DeckDelete from './deck-delete';
 import { initializeDeckBuilder } from '../lib/deck-utils';
+import UserContext from '../components/user-context';
 
 export const deckCardsQuery = gql`
   query($id: Int!) {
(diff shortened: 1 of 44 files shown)


next/components/deck.js
```diff
@@ -37,32 +41,33 @@ const getDeckToExport = (deckCards, deckName, path = null, power = null) => {
...
+    <>
+      <h1 className="deckName">{deck.name}</h1>
+      <h2>Power</h2>
+      {power ? power.name : 'No Power Selected'}
+      <h2>Path</h2>
+      {path ? path.name : 'No Path Selected'}
+      <h2>Cards</h2>
+      <DeckCardList deckCards={cards} />
+      <DeckExport deckInProgress={deckToExport} />
+      {user && user.id === deck.authorId && <DeckDelete deck={deck} />}
```
lily-pytel: I personally prefer to let the component do this kind of logic - as in give `DeckDelete` the user context and let it decide whether it should return null or content. I'm not a fan of these conditional statements in components but full disclosure, I did not research whether this is a performance concern or something. I just recently read that `Boolean(thing)` is apparently 91% slower than `!!thing` so... /shrug

jtrussell: Sure thing. Honestly, my preference would have been to go the complete other direction and  put the logic in the page so happy to get out of the middle ground.

lily-pytel: Also this is what I see when I examine `user`:
image

Which would mean you'd have to do `user.user.id` but I also got null for `deck.authorId` for a deck I just created so I'm thinking I'm not testing this right

lily-pytel: Might be because `save-deck` doesn't seem to save a deck with an author so do we actually have a way for users to save deck or is that out of scope?

jtrussell: Ah, good catch. I messed up and forgot to give the context value its princess Leia hair `{   }`. Should be fixed! I also moved the "are we doing delete things?" logic to the `DeckDelete` component - good suggestion 👍.

lily-pytel:
Looks great! But I reserve judgement until I can test, which should happen after this is merged: https://github.com/mythgard-hub/mythgard-hub/pull/61

lily-pytel:
@jtrussell unblocker merged! Resolve yo conflicts

jtrussell:
Conflicts resolved.

jtrussell:
I hear ya, I couldn't think of a more elegant way to make sure only one of the conditions ran.

Finding zero decks (e.g. when one is deleted) results in no errors but also gives us nothing we can render with the deck components.

PR #59, opened 2019-08-31
Cards paths powers tabs (part 1 - all form, no function)

alexmiddeleer:
I left the tabs as real buttons since it's fairly a11y friendly I'm assuming (that's why there's an orange outline if anyone's wondering, easy keyboard nav). If we think we need some of the fancier aria attributes I'm fine to put those in, but I don't know how world-changing they actually are.

### DIFF
--- a/next/components/tab-group.js
+++ b/next/components/tab-group.js
@@ -1,0 +1,48 @@
+import React, { useState, useEffect } from 'react';
+import PropTypes from 'prop-types';
+
+function TabGroup({ onChange, name, labels }) {
+  const [selectedLabel, setLabel] = useState(labels[0]);
+
+  useEffect(() => onChange(selectedLabel));
+
+  return (
+    <div data-cy={`tabs_${name}`}>
+      <style jsx>{`
+        display: flex;
+        flex-wrap: none;
+        border-bottom: 1px solid orange;
+        .tab {
+        }
+        .tab-spacer {
+          min-width: 20px;
+        }
+      `}</style>
+      <div className="tab-spacer"></div>
+      {labels.map((label, i) => {
+        return (
+          <>
+            <button
+              name={`${label} tab`}
+              key={2 * i}
+              className="tab reset-button"
+              onClick={() => setLabel(label)}
+            >
+              {label}
+            </button>
+            <div key={2 * i + 1} className="tab-spacer"></div>
+          </>
+        );
+      })}
+      <div className="tab-spacer"></div>
+    </div>
+  );
+}
+
+TabGroup.propTypes = {
+  onChange: PropTypes.func.isRequired,
+  name: PropTypes.string.isRequired,
+  labels: PropTypes.array.isRequired
+};
+
+export default TabGroup;
--- a/next/components/tab-group.js
+++ b/next/components/tab-group.js
@@ -7,15 +7,35 @@ function TabGroup({ onChange, name, labels }) {
   useEffect(() => onChange(selectedLabel));
 
   return (
-    <div data-cy={`tabs_${name}`}>
+    <div className="mg-tab-group" data-cy={`tabs_${name}`}>
       <style jsx>{`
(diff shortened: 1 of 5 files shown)


next/components/tab-group.js
```diff
@@ -0,0 +1,79 @@
+import React, { useState, useEffect } from 'react';
+import PropTypes from 'prop-types';
+import { ThemeContext } from './theme-context';
+
+function TabGroup({ onChange, name, labels }) {
+  const [selectedLabel, setLabel] = useState(labels[0]);
+
+  useEffect(() => onChange(selectedLabel));
```
lily-pytel: Placeholders for future functionality?

next/components/tab-group.js
```diff
@@ -0,0 +1,79 @@
...
+
+  return (
+    <ThemeContext.Consumer>
+      {theme => (
+        <div className="mg-tab-group" data-cy={`tabs_${name}`}>
+          <style jsx>{`
+            .mg-tab-group {
+              display: flex;
+              flex-wrap: nowrap;
+              font-size: 11px;
```
lily-pytel: The mock has the font a little bigger (13.3px though I would just go with 14px): https://mythgard-hub.github.io/deckb_template.html

lily-pytel:
Though I admit I hate that orange outline when you click it.

next/components/tab-group.js
alexmiddeleer: yep

next/components/tab-group.js
alexmiddeleer: For me it shows as 11px when I inspect it in Chrome..

alexmiddeleer:
> Though I admit I hate that orange outline when you click it.

I don't think that part is great either. Open to suggestions.

next/components/tab-group.js
lily-pytel: That's so weird! I do think 11px is kind of small in general but it's obviously small potatoes

lily-pytel:
> I don't think that part is great either. Open to suggestions.

How about just making it divs?

alexmiddeleer:
> > I don't think that part is great either. Open to suggestions.
> 
> How about just making it divs?

It would still need some sort of outline and focus-ability, otherwise it can't be tab-navigated which is not accessible design. I think I'll remove the orange outline though and just let Chrome leave in its standard, nearly unnoticeable blue halo.

jtrussell:
Another option would be to use links for tabs instead of buttons so each tab would get its own route. Added benefit: we could link directly to the paths/powers view of a decklist.

This might also push us toward tracking deck modifications in local storage like Lily was getting at [here](https://trello.com/c/wCR2rE1O/125-deck-builder-save-deck-when-user-is-not-logged-in#comment-5d6c7dff2b03901f6310fb54).

lily-pytel:
I am super in favor of each tab getting its own route in general!

jtrussell:
Agreed! The only awkward thing is Next doesn't give us a concept of "sub" routes out of the box (at least not that I'm aware of).

We'd have to do something like:

```
(Folders)
next/
  pages/
    deck-builder/
      index.js
      cards.js
      paths.js
      powers.js
```

Where `deck-builder/index.js` might render everything but the tab content and `deck-builder/paths.js` would do something like:

```javascript
import BuilderShell from './index'

export default function PathTab() {
  return (
    <BuilderShell>
      Path Tab!
    </BuilderShell>
  )
}
```

jtrussell:
Actually, I take that back. It looks like there is some fancy dynamic routing stuff we could use!

https://github.com/zeit/next.js#dynamic-routing

alexmiddeleer:
Oh wow, I didn’t even think of making the tabs change the url. Now I want
to make all of the filters part of the url 🤭.

jtrussell:
I like it! Though don't feel compelled to overhaul before with v1 bearing down ;).

PR #61, opened 2019-09-01
Save deck with author

lily-pytel:
- Many components lost the `<Query>` component. 👋 
- Many components lost their inline queries. They have been added to `deck-queries.js`. I think that file should be moved to the `queries` folder, but that seems better suited for another PR
- Decks get saved with their author
- Deck can no longer be saved if there is no logged in user. I will create a separate PR to bring them to the login page and stuff but this unblocks @jtrussell's delete deck stuff.
- Author added to the deck page
- Author added to the deck list page

### DIFF
--- a/e2e/cypress/integration/deck-builder-page.js
+++ b/e2e/cypress/integration/deck-builder-page.js
@@ -109,11 +109,13 @@ describe('Deck builder page', () => {
       .then(cards => {
         expect(numCardsBeforeFilter).to.equal(cards.length);
 
-        // finally save the deck
+        // finally try to save the deck and fail
         cy.get('[data-cy="deckTitle"]').type('Floop the Pig');
         cy.get('[data-cy="saveDeck"]').click();
-        return cy.location().should(location => {
-          expect(location.pathname).to.eq('/deck');
+        cy.on('window:alert', str => {
+          expect(str).to.equal(
+            'You are not logged in. You will be dealt with in another PR'
+          );
         });
       });
   });
--- a/next/lib/mutations/add-deck.js
+++ b/next/lib/mutations/add-deck.js
@@ -1,13 +1,29 @@
 import gql from 'graphql-tag';
 
 const addDeckMutation = gql`
-  mutation AddDeck($name: String!, $pathId: Int, $powerId: Int) {
+  mutation AddDeck(
+    $name: String!
+    $pathId: Int
+    $powerId: Int
+    $authorId: Int
+  ) {
     createDeck(
-      input: { deck: { name: $name, pathId: $pathId, powerId: $powerId } }
+      input: {
+        deck: {
+          name: $name
+          pathId: $pathId
+          powerId: $powerId
+          authorId: $authorId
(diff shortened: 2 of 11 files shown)


e2e/cypress/integration/deck-builder-page.js
```diff
@@ -109,11 +109,13 @@ describe('Deck builder page', () => {
       .then(cards => {
         expect(numCardsBeforeFilter).to.equal(cards.length);
 
-        // finally save the deck
+        // finally try to save the deck and fail
```
lily-pytel: Temporary. The next PR will check that you were directed to the log in page. I think I will stop the test there since I don't know how to test that a user has actually logged in

alexmiddeleer: Probably by stubbing the request that tells the app if the user is logged in. https://docs.cypress.io/guides/guides/network-requests.html#Stubbing. Don't know if SSR adds any complexity but hopefully not.

We should probably write a login and signup test that doesn't get stubbed, though.

next/lib/mutations/add-deck.js
```diff
@@ -30,7 +46,8 @@ const createNewEmptyDeck = (apolloClient, deck) => {
     variables: {
       name: deck.deckName,
       pathId: path,
-      powerId: power
+      powerId: power,
+      authorId
```
alexmiddeleer: Just to confirm, this will throw an error if there is no account with that id in the db, right?

lily-pytel:
> The refactor looks awesome, hooks are much cleaner and I like having the queries in a lib. Though in the future when doing a big refactor I do recommend splitting the PRs so you have the "this changes nothing but makes the code nicer to read" separate from the "shit's new" parts. Helps with understanding the intent of the changes.

@alexmiddeleer I always want to do this and end up doing the refactor while doing the other stuff. Sorry!

next/lib/mutations/add-deck.js
lily-pytel: Correct but that error doesn't get displayed anywhere at the moment. I need to either learn how to use the `ApolloConsumer` component and try to use `useMutation` on these as well, but seemed like something for another day

e2e/cypress/integration/deck-builder-page.js
lily-pytel: And log in to some fake google account on Travis?

alexmiddeleer:
> > The refactor looks awesome, hooks are much cleaner and I like having the queries in a lib. Though in the future when doing a big refactor I do recommend splitting the PRs so you have the "this changes nothing but makes the code nicer to read" separate from the "shit's new" parts. Helps with understanding the intent of the changes.
> 
> @alexmiddeleer I always want to do this and end up doing the refactor while doing the other stuff. Sorry!

Best approach there (imo) is you code a bit, realize you're horrified with the code you're touching, switch back to master, make a refactor branch, and then when you're done refactoring rebase the code changes branch back onto the refactor branch.

PR #63, opened 2019-09-02
Restrict account updates to authors

evsheffield:
Sorry that this one was a long time coming. It took some time for me to figure out how to get things working but fortunately I've learned a lot along the way.

First off, some things to be aware of:

- There are updates to the `.env` that need to be made. You can see the changes in the template file.
- There is a new node module dep in the `next` project.

I can add a wiki page to document some of this if it would help, but roughly how this works is:

- The JWT cookie (when available) is sent as an `Authorization` request header for all GraphQL requests
- Postgraphile verifies and reads the auth header and extracts all of the claims in it as properties. It also reads the "role" claim to set a Postgres role.
- I've created three high-level postgres roles: `admin` (for future use, nothing sets this role atm), `authd_user` (for logged in users), and `anon_user` (for non-authenticated users). You can grant or deny various permissions to the DB schema, tables, etc. to each of these roles. In this PR, only admins and logged in users can update the account table.
- Row level policies are put in place so that users can only update the account which they are currently logged in as. These policies can read the claims made in the JWT using `current_setting`.

Some gotchas that might be good to be aware of:

- Table owners generally bypass all row level security policies. This is why Postgraphile now logs in to the DB with a different user than the one that creates the DB.
- As soon as you enable row level security on a table, it assumes a default deny everything policy, so you will need to make sure the needed roles have the appropriate privileges to even be able to do anything including read any row.

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -21,6 +21,9 @@ PGDATABASE=postgres
 PGPORT=5432
 PGSSL=no
 
+EXPRESS_PGUSER=postgres
+EXPRESS_PGPASSWORD=example
+
 # Configs For JWTs
 JWT_COOKIE_NAME=jwt
 JWT_SECRET=banana_cat_fondue
--- a/.env.template
+++ b/.env.template
@@ -14,9 +14,9 @@ GA_TRACKING_ID=XX-YYYYYYYYY-1
 MG_CDN=https://cdn.mythgardhub.com
 
 # PostgreSQL connection info
-PGUSER=postgres
+PGUSER=postgraphile
 PGHOST=db
-PGPASSWORD=example
+PGPASSWORD=bears4life
 PGDATABASE=postgres
 PGPORT=5432
 PGSSL=no
--- a/.env.template
+++ b/.env.template
@@ -24,6 +24,10 @@ PGSSL=no
 EXPRESS_PGUSER=postgres
 EXPRESS_PGPASSWORD=example
 
+# Roles assumed by Postgraphile for making DB requests
+PG_AUTHD_USER_ROLE=authd_user
+PG_ANON_USER_ROLE=anon_user
+
 # Configs For JWTs
 JWT_COOKIE_NAME=jwt
 JWT_SECRET=banana_cat_fondue
--- a/next/components/init-apollo.js
+++ b/next/components/init-apollo.js
@@ -1,5 +1,6 @@
 import { ApolloClient, InMemoryCache } from 'apollo-boost';
 import { BatchHttpLink } from 'apollo-link-batch-http';
+import { setContext } from 'apollo-link-context';
 import fetch from 'isomorphic-unfetch';
 
 let apolloClient = null;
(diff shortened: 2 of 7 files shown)


postgres/init.sql
```diff
@@ -156,3 +178,29 @@ $$ language 'plpgsql';
 CREATE TRIGGER update_deck_modtime
   BEFORE UPDATE ON mythgard.deck
   FOR EACH ROW EXECUTE PROCEDURE update_modified_column();
+
+-- Create a user for Postgraphile to connect as which has permissions
+-- to the mythgard schema
+
+CREATE USER postgraphile WITH password 'bears4life';
```
evsheffield: Naturally this PW will need to not be here for production. If anyone knows how we could reference the .env variable value for this I can get that taken care of.

.env.template
```diff
@@ -14,13 +14,20 @@ GA_TRACKING_ID=XX-YYYYYYYYY-1
...
-PGUSER=postgres
+PGUSER=postgraphile
 PGHOST=db
-PGPASSWORD=example
+PGPASSWORD=bears4life
 PGDATABASE=postgres
 PGPORT=5432
 PGSSL=no
 
+EXPRESS_PGUSER=postgres
```
alexmiddeleer: This is basically an admin, right? Perhaps we should restrict the express server as well, unless the cost/benefit is not worth it.

next/components/init-apollo.js
```diff
@@ -17,17 +18,41 @@ const makeCache = initialState =>
...
+  let token;
+  const jwtCookie = cookies.find(cook =>
+    cook.startsWith(`${process.env.JWT_COOKIE_NAME}=`)
+  );
+  if (jwtCookie) {
+    token = jwtCookie.split('=')[1];
+  }
+  const newHeaders = { ...headers };
+  if (token) {
+    newHeaders.Authorization = `Bearer ${token}`;
```
jtrussell: Bearer indeed ;)

next/components/init-apollo.js
```diff
@@ -17,17 +18,41 @@ const makeCache = initialState =>
...
-      // Use fetch() polyfill on the server
-      fetch: !process.browser && fetch
-    }),
+    link: authLink.concat(
+      new BatchHttpLink({
+        batchMax: 200,
+        uri,
+        // credentials: "same-origin", // Additional fetch() options like `credentials` or `headers`
+        // Use fetch() polyfill on the server
+        fetch: !process.browser && fetch
```
jtrussell: Huh, mental note to check whether `process.browser` is ever - and move the import for isomorphic-unfetch down here. IIRC we instructed webpack to not include the isomorphic-unfetch module in the bundles that get shipped to the browser. Not a problem for this line per-se, but the `import` at the top of the file will fail if we try to `import` iso-unfetch in the browser.

postgres/init.sql
jtrussell: Untested, but perhaps using a `PGOPTIONS` env variable + `current_setting` will work:
https://dba.stackexchange.com/questions/52235/how-can-i-use-an-environment-variable-in-a-postgres-function

next/components/init-apollo.js
alexmiddeleer: I believe I fetched this (🚎 )  from the example here https://github.com/zeit/next.js/blob/master/examples/with-apollo/lib/init-apollo.js

Not sayin' they're not 🤡 s though. Cookbooks are a step down from docs, and we all know to never trust the docs 100%.

jtrussell: Fair enough. It's a trivial concern.

For reference to the curious though, the docs link to [this page](https://arunoda.me/blog/ssr-and-server-only-modules) which goes into to detail on the topic of SSR only imports. Ctrl-f [the docs](https://nextjs.org/docs) for "import them properly"). Looking at the prod bundle I think the node-fetch stuff (via iso unfetch) is making it in there.

alexmiddeleer:
Thanks Justin, excellent read.

.env.template
evsheffield: Yes, since this user is the "owner" of everything they basically hold all the keys to the kingdom. It certainly makes sense to tighten up their privileges. I think the express server currently only needs to interact with the accounts table so we can restrict it accordingly. :+1:

alexmiddeleer:
That could be a subsequent pr if it makes sense to break them up.

postgres/init.sql
evsheffield: Thanks! Unfortunately I was unsuccessful in some brief fiddling with it since it seems not to like `current_setting` in a `CREATE USER` but it seems like a potential path with some more investigation. To keep things moving along I'm going to punt it to https://trello.com/c/AWdFQCCA/6-productionize-database for now.

PR #64, opened 2019-09-03
Paging (Replace card names with images step 1)

alexmiddeleer:
Thinking of just skipping the unit tests for the paging lib for now. Bugs would be pretty easy to spot in the widget.

Right now all paging is client-side. I think this is fine/preferable for cards. It should be easy to upgrade the paging component to support a bubbling click event, as the paging component doesn't manage which items to show or how data is loaded.

### DIFF
--- a/next/components/card-list.js
+++ b/next/components/card-list.js
@@ -1,3 +1,4 @@
+import React from 'react';
 import PropTypes from 'prop-types';
 import CardListItem from './card-list-item';
 
@@ -12,6 +13,7 @@ export default function CardList({ onCardClick, cards }) {
     </ul>
   );
 }
+
 CardList.propTypes = {
   cards: PropTypes.array,
   onCardClick: PropTypes.func
--- a/next/components/card-list.js
+++ b/next/components/card-list.js
@@ -36,20 +36,21 @@ export default function CardList({ onCardClick, cards, pageSize }) {
             max-width: 100px;
           }
         `}</style>
-        {showPrev && (
-          <button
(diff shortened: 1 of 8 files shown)


next/components/card-list.js
```diff
@@ -1,18 +1,41 @@
...
-      ))}
-    </ul>
+    <>
+      <ul className="cardList" data-cy="cardList">
+        {cards.map((card, index) => {
+          if (!onCurrentPage(index, currentPage, pageSize)) {
+            return;
+          }
+          return (
+            <li key={card.id ? card.id : index}>
```
jtrussell: I think `0` is a valid card id.

next/components/paging-controls.js
```diff
@@ -0,0 +1,63 @@
...
+    </div>
+  );
+}
+
+PagingControls.defaultProps = {
+  pageSize: 20
+};
+
+PagingControls.propTypes = {
+  currentPage: PropTypes.any.isRequired,
```
evsheffield: I think the first three props here should be `PropTypes.number`

next/lib/paging.js
```diff
@@ -0,0 +1,25 @@
+// 1, 20 => 20
```
evsheffield: I think this comment differs from the output of the function. Seems like the first page is indexed as 0, is that right?

alexmiddeleer: Oops, yeah it should be 0, 20 => 20. The output is not zero-based but current page is. I updated the comment.

PR #65, opened 2019-09-05
Card images

alexmiddeleer:
Please forgive the jankiness of the image optimization script. It was kind of a pain in the ass.

### DIFF
--- a/next/lib/card.js
+++ b/next/lib/card.js
@@ -1,0 +1,22 @@
+export const nameToImage = name => {
+  name = name
+    .trim()
+    .toLowerCase()
+    .replace(/[^0-9a-z -]/gi, '');
+  if (!name.length) {
+    return '';
+  }
+  return name
+    .split(' ')
+    .map(s => {
+      const rest = s.length > 1 ? s.slice(1) : '';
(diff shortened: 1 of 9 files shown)


next/lib/card.js
```diff
@@ -0,0 +1,32 @@
...
+  return name
+    .split(' ')
+    .map(s => {
+      const rest = s.length > 1 ? s.slice(1) : '';
+      return `${s[0].toUpperCase()}${rest}`;
+    })
+    .join('_')
+    .split('-')
+    .map(s => {
+      const rest = s.length > 1 ? s.slice(1) : '';
```
evsheffield: Since this appears to be the same function used with the previous map you could consider factoring it out and reusing the function.

next/lib/card.js
```diff
@@ -0,0 +1,32 @@
...
+      return `${s[0].toUpperCase()}${rest}`;
+    })
+    .join('-');
+};
+
+export const imagePath = (name, set = 'core') => {
```
evsheffield: Maybe this could be called `imagePathSmall` to be more specific about which path it's getting.

PR #66, opened 2019-09-05
Deck list styles + results table

lily-pytel:
Looks like this: 

image

Mockup: https://mythgard-hub.github.io/decks_template.html

## Musings

- There are more columns in that table... what are they? Are they v1? Is one of them total mana cost or upvotes or what?
- Looks like the creator is a link but we don't have plans for a user page in v1... right? Only your own profile page.
- There are other filters in the mock that are also not included because I don't understand them. There are cards in the todo column in trello for them. 
- The way I get the factions for each deck feels so gross, but I could not find proper documentation on the `group` argument\variable\function in graphQL. It almost feels like it's not real. This is basically the only thing I've found so far and it didn't work for me using all kinds of combinations: https://www.gatsbyjs.org/docs/graphql-reference/#group. In the end, I basically did something that feels like CS102 so... open to ideas.

### DIFF
--- a/next/components/deck-list.js
+++ b/next/components/deck-list.js
@@ -6,22 +6,7 @@ import { ThemeContext } from './theme-context';
 import { deckFactions } from '../lib/deck-queries';
 import { FACTION_IMAGES } from '../constants/factions';
 import ErrorMessage from './error-message';
-
-function formatDataToFactions(data) {
-  try {
-    return data.cardDecks.nodes.reduce((factionsPerDeck, c) => {
-      const factions = c.card.cardFactions.nodes.map(f => f.faction.name);
-
-      factionsPerDeck[c.deckId] = factionsPerDeck[c.deckId]
-        ? new Set([...factionsPerDeck[c.deckId], ...factions])
-        : factions;
-
-      return factionsPerDeck;
-    }, {});
-  } catch (e) {
-    return {};
-  }
-}
+import { collectDeckFactions } from '../lib/deck-utils';
 
 export default function DeckList({ decks }) {
   if (!decks || !decks.length) return null;
@@ -38,7 +23,7 @@ export default function DeckList({ decks }) {
   if (loading) return 'Loading...';
   if (error) return <ErrorMessage message={error.message} />;
 
-  const factions = formatDataToFactions(data);
+  const factions = collectDeckFactions(data);
 
   return (
     <div>
--- a/next/components/deck-list.js
+++ b/next/components/deck-list.js
@@ -1,23 +1,121 @@
+import { useContext } from 'react';
 import Link from 'next/link';
 import PropTypes from 'prop-types';
+import { useQuery } from 'react-apollo-hooks';
+import { ThemeContext } from './theme-context';
+import { deckFactions } from '../lib/deck-queries';
+import { FACTION_IMAGES } from '../constants/factions';
+import ErrorMessage from './error-message';
+
+function formatDataToFactions(data) {
+  try {
+    const factionsPerDeck = {};
+
+    data.cardDecks.nodes.forEach(c => {
+      const factions = c.card.cardFactions.nodes.map(f => f.faction.name);
+
+      factionsPerDeck[c.deckId] = factionsPerDeck[c.deckId]
+        ? [...factionsPerDeck[c.deckId], ...factions]
+        : factions;
+    });
+
+    return factionsPerDeck;
+  } catch (e) {
+    return {};
+  }
+}
 
 export default function DeckList({ decks }) {
-  return (
-    <ul>
-      {decks.map((deck, index) => {
-        const author = deck && deck.author ? deck.author.username : 'unknown';
+  if (!decks || !decks.length) return null;
 
-        return (
-          <li data-cy="deckListItem" key={deck.id}>
-            <Link href={`/deck?id=${deck.id}`} key={index}>
-              <a>
-                {index + 1}. name: {deck.name}. by: {author}
-              </a>
-            </Link>
-          </li>
-        );
-      })}
-    </ul>
+  const theme = useContext(ThemeContext);
+
+  const deckIds = decks.map(deck => deck.id);
+  const { loading, error, data } = useQuery(deckFactions, {
+    variables: {
+      deckIds
+    }
+  });
+
+  if (loading) return 'Loading...';
+  if (error) return <ErrorMessage message={error.message} />;
+
+  const factions = formatDataToFactions(data);
+
+  return (
+    <div>
+      <style jsx>{`
+        .deckListTable {
+          width: 100%;
+          border-collapse: collapse;
+        }
+        .deckListRow {
+          border-top: ${theme.border};
+        }
+        .deckListRow td {
+          padding: 15px 10px 15px 10px;
+        }
+        .zebraRow {
+          background-color: ${theme.zebraRowBackroung};
+        }
+        img {
+          max-height: 15px;
+          vertical-align: top;
+          margin-right: 5px;
+        }
+        .deckName :global(a) {
+          font-size: 18px;
+          font-weight: bold;
+          text-decoration: none;
+        }
+        .deckAuthor {
+          font-size: 14px;
+          line-height: 2;
+        }
+        .modifiedDate {
+          float: right;
+        }
+        .factions {
+          text-align: center;
+        }
+      `}</style>
+      <table className="deckListTable">
+        <tbody>
+          {decks.map((deck, index) => {
+            const classNames = `deckListRow ${index % 2 ? 'zebraRow' : ''}`;
+            const author =
+              deck && deck.author ? deck.author.username : 'unknown';
+            const modified = new Date(deck.modified);
+            const currFactions = factions[deck.id]
+              ? [...new Set(factions[deck.id])].map(f => (
+                  <img src={FACTION_IMAGES[f]} />
+                ))
+              : '';
+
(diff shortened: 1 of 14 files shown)


next/components/deck-list.js
```diff
@@ -1,23 +1,121 @@
+import { useContext } from 'react';
 import Link from 'next/link';
 import PropTypes from 'prop-types';
+import { useQuery } from 'react-apollo-hooks';
+import { ThemeContext } from './theme-context';
+import { deckFactions } from '../lib/deck-queries';
+import { FACTION_IMAGES } from '../constants/factions';
+import ErrorMessage from './error-message';
+
+function formatDataToFactions(data) {
```
lily-pytel: I am so ashamed of this. Basically achieves the equivalent of a `group by` clause in SQL. It takes the data from the card deck factions nested-to-hell query and attempts to, ya know, group by deck id. KMD, as Shelley would say.

next/components/layout.js
```diff
@@ -33,10 +33,21 @@ function Layout({ title, desc, children }) {
           margin: 0;
         }
 
+        h1 {
```
lily-pytel: This does change some things in other pages but those are unpolished pages style-wise

next/components/page-banner.js
```diff
@@ -16,12 +16,6 @@ function PageBanner({ children, image }) {
           flex-direction: row-reverse;
           align-items: center;
         }
-        .page-banner h1 {
```
jtrussell: We might want to keep some of the `.page-banner h1` styles here rather than make them global. E.g. the margin and padding. It makes sense to move out color to Layout though.

next/components/deck-list.js
```diff
@@ -1,23 +1,121 @@
...
+  try {
+    const factionsPerDeck = {};
+
+    data.cardDecks.nodes.forEach(c => {
+      const factions = c.card.cardFactions.nodes.map(f => f.faction.name);
+
+      factionsPerDeck[c.deckId] = factionsPerDeck[c.deckId]
+        ? [...factionsPerDeck[c.deckId], ...factions]
```
evsheffield: I may totally be missing something but does this end up giving you duplicate factions in the list if you have multiple cards of the same faction?

next/components/page-banner.js
lily-pytel: Truth. Fixed.

next/components/deck-list.js
```diff
@@ -1,23 +1,121 @@
...
+      `}</style>
+      <table className="deckListTable">
+        <tbody>
+          {decks.map((deck, index) => {
+            const classNames = `deckListRow ${index % 2 ? 'zebraRow' : ''}`;
+            const author =
+              deck && deck.author ? deck.author.username : 'unknown';
+            const modified = new Date(deck.modified);
+            const currFactions = factions[deck.id]
+              ? [...new Set(factions[deck.id])].map(f => (
```
lily-pytel: @evsheffield to your question, the other line would give duplicates and this line would get rid of them. Do you know of a cleaner way to only add new values to an array?

next/components/deck-list.js
lily-pytel: Nope, you're right, and they get smashed here: https://github.com/mythgard-hub/mythgard-hub/pull/66/files#diff-3d7d79e20a9f2b29e37a3a39f804234aR90. I could move the smashing to the same function to make it clearer

lily-pytel: And if we don't care about the performance of a ~250 array (that's the max deck size, right?), this can be changed to `new Set([...factionsPerDeck[c.deckId], ...factions])`. What do you think?

evsheffield: Cool, that's the piece I was missing. Not a big deal but I think it might be nice to have the smashing in the same function. Rather than make a new set each time I suppose you could `add` each of the new elements, I'm not sure how the performance would compare. But either way that seems good to me! 

```
if (!factionsPerDeck[c.deckId]) {
  factionsPerDeck[c.deckId] = new Set();
}
factions.forEach(fac => {
  factionsPerDeck[c.deckId].add(fac);
});
```

alexmiddeleer: I don't think you need to be ashamed of it, but I do think I would have used `reduce`. NBD though, we can refactor later.

alexmiddeleer: Might move this to a lib and improve naming. This api seems clean to me:

```
import { collectDeckFactions } from '../lib/deck-queries.js'
... Get data, do stuff.
const factions = collectDeckFactions(deckData);
```

next/components/deck-list.js
```diff
@@ -1,23 +1,121 @@
...
+      const factions = c.card.cardFactions.nodes.map(f => f.faction.name);
+
+      factionsPerDeck[c.deckId] = factionsPerDeck[c.deckId]
+        ? [...factionsPerDeck[c.deckId], ...factions]
+        : factions;
+    });
+
+    return factionsPerDeck;
+  } catch (e) {
+    return {};
```
alexmiddeleer: we should log the error. might need a logging lib. Can simply be a console.error wrapper for now.

next/components/deck-list.js
alexmiddeleer: For this you could use an object like a bitmap e.g. `{ oberos: true, norden: false}` and then whenever you encounter a faction set that faction to true. Then at the end just pull out the trues with Object.entries().reduce().

next/components/layout.js
```diff
@@ -33,10 +33,19 @@ function Layout({ title, desc, children }) {
...
+          color: ${theme.fontColorSelected};
```
alexmiddeleer: The selected color should not be coupled to the heading color. We should make a new prop `${theme.fontColorH1}` or `${theme.fontColorHeading}` if you prefer.

next/components/layout.js
```diff
@@ -48,10 +57,31 @@ function Layout({ title, desc, children }) {
...
+          padding: 10px 10px;
           width: 100%;
           border-radius: 10px;
+          font-style: italic;
+        }
+
+        input[type='text'],
+        select {
+          height: 30px;
+          border: ${theme.tabBorder};
```
alexmiddeleer: tabBorder is just for tabs. I recommend an inputBorder field. The goal is to be able to change the tabBorder color without having to worry about changing things that aren't tabs accidentally or being forced to touch code all over the app.

next/components/theme-context.js
```diff
@@ -9,14 +9,17 @@ export const themes = {
...
     tabColor: mgColors.lightBlue,
     border: `1px solid ${mgColors.orange}`,
     fontFamily: mgFonts.exo2,
     buttonOutline: `1px solid ${mgColors.orange}`,
     buttonOutlineOffset: '6px',
-    buttonBorderDisabled: `2px solid ${mgColors.lightGray}`
+    buttonBorderDisabled: `2px solid ${mgColors.lightGray}`,
+    inputBackground: mgColors.lightGray,
+    inputPlaceholderTextColor: mgColors.lightBlue,
+    zebraRowBackroung: mgColors.grayBlue
```
alexmiddeleer: `zebraRowBackroung` spelling

next/components/deck-list.js
```diff
@@ -1,23 +1,121 @@
...
+  }
+}
 
 export default function DeckList({ decks }) {
+  if (!decks || !decks.length) return null;
+
+  const theme = useContext(ThemeContext);
+
+  const deckIds = decks.map(deck => deck.id);
+  const { loading, error, data } = useQuery(deckFactions, {
```
alexmiddeleer: I don't quite understand why we are making another query to the database in this component. Shouldn't we amend the query in the parent to return the factions so that we're not making another network request?

next/components/deck-list.js
lily-pytel: @alexmiddeleer I'm not a huge fan of the reduce version, personally. What do you think? 
16bdc1a06e3f737c874b18dff9c8639fa99d9eab

(I did fix console.log to console.error, coming up)

next/components/deck-list.js
lily-pytel: Honestly didn't know that was a thing until this comment and looking closely at the `deckCardsQuery`. That's much better! Fixed

next/components/layout.js
lily-pytel: True

next/components/layout.js
lily-pytel: I guess I saw it more as a color palette than colors for specific sections but you're right.

next/components/theme-context.js
lily-pytel: How did the g get all the way back there

next/components/deck-list.js
lily-pytel: Only reason I wouldn't do this is in case the factions change. I'd like to keep it as agnostic to the game as possible. What do you think?

lily-pytel:
@alexmiddeleer I believe I addressed all your notes except for the one about the bitmap. I can do that one too though

next/components/deck-list.js
alexmiddeleer: This comment isn't one of the important ones, so no need to keep refactoring as there's better things to do. But since it's interesting to discuss: you could construct it on the fly, e.g. start with `const bitmap = {}` and while you loop through, you set `bitmap[factionName] = true`. Then you can pull out the entries, remove the ones that are false, and those are the factions. Though OTOH we have other places in the code that rely on hardcoded factions, like the icon images const. /shrug :)

PR #68, opened 2019-09-06
Include articles from TRS

jtrussell:
I'm not signifying an intent to merge... just wafting this out there so you folks can react to it and tell me which ideas are most horrible. FWIW I imagine the first incarnation of articles to be temporary no matter what.


jtrussell:
Screen Shot 2019-09-05 at 10 37 59 PM

And here's an example of whit it might look like.

jtrussell:
Screen Shot 2019-09-05 at 11 14 06 PM

And here's an example of what it might look like.

next/components/article-list.js
```diff
@@ -0,0 +1,54 @@
...
+
+// Kinda janky, but this is all just kinda janky
+const CATEGORY_MYTHGARD = 'Mythgard';
+
+export default function ArticleList() {
+  const [list, setList] = useState([]);
+  const [error, setError] = useState(null);
+  const [isLoading, setIsLoading] = useState(true);
+
+  useEffect(() => {
```
alexmiddeleer: Not sure what useEffect affords us here. A fancy way to run an async function inside a render function?

next/components/article.js
```diff
@@ -0,0 +1,80 @@
...
+    timeMsg = 'Over a year ago';
+  } else if (monthsDelta > 1) {
+    timeMsg = `${monthsDelta} months ago`;
+  } else if (daysDelta > 1) {
+    timeMsg = `${daysDelta} days ago`;
+  } else if (hrsDelta > 1) {
+    timeMsg = `${hrsDelta} hours ago`;
+  } else {
+    timeMsg = 'Just now';
+  }
```
alexmiddeleer: You were very charitable not to just plop in the publish date here as a string :P

next/components/article.js
```diff
@@ -0,0 +1,80 @@
...
+
+        .article-link {
+          color: ${theme.font};
+        }
+
+        .article-date {
+          font-style: italic;
+        }
+
+        .article-divider {
```
alexmiddeleer: looks fancy. is this gradient something we can add to the theme or is it pretty article-specific?

next/components/article.js
```diff
@@ -0,0 +1,80 @@
...
+        }
+      `}</style>
+      <div className="article">
+        <a className="article-link" href={url}>
+          {title}
+        </a>
+        <div className="article-date">{timeMsg}</div>
+        <hr className="article-divider" />
+        <div className="article-desc">
+          <div dangerouslySetInnerHTML={{ __html: description }} />
```
alexmiddeleer: hmm... what could go wrong 😄

next/components/article-list.js
```diff
@@ -0,0 +1,54 @@
...
+  if (error) return <ErrorMessage message="Unable to load articles." />;
+  if (isLoading) return <div>Loading</div>;
+  if (!list.length) return <div>No recent articles.</div>;
+
+  return (
+    <div>
+      {list.map((feedItem, ix) => (
+        <Article
+          key={ix}
+          title={feedItem.title}
```
alexmiddeleer: all of these props should probably have name dangerousTitle, dangerousUrl, etc.?

next/server-routes/articles.js
```diff
@@ -0,0 +1,8 @@
+const request = require('request');
+const router = require('express').Router();
+
+router.get('/rss', async (req, res) => {
+  request('https://teamrankstar.com/feed').pipe(res);
```
alexmiddeleer: Do we really need this or can we force it to be client side?

next/components/article-list.js
jtrussell: It simulates a "component did mount" lifecycle method. It'll run the first time the component is mounted to the page, not before, and only once.

next/components/article.js
jtrussell: iirc the was one other spot we had a similar gradient but it's not universal.

next/server-routes/articles.js
jtrussell: CORS

jtrussell: The TRS folks would have to allow it.

jtrussell:
Agree with you on the XSS front, not super worried about it for this poc/v1 implementation though since "we" effectively own the content getting pulled in.  ¯\_(ツ)_/¯

jtrussell:
Still a moot point until/unless the TRS feed gets cleaned up. My preference remains for static, curated content and figuring out how we do more later.

alexmiddeleer:
👍

jtrussell:
Shutting this down.

PR #69, opened 2019-09-07
Tabs paths powers (pr 2)

alexmiddeleer:
Sorry about the somewhat bloated PR.


next/lib/queries/powers-query.js
```diff
@@ -1,11 +1,12 @@
 import gql from 'graphql-tag';
 
 const allPowersQuery = gql`
-  query pathsAndPowers {
```
lily-pytel: My bad, bad copy-pasting

next/lib/queries/paths-query.js
```diff
@@ -1,11 +1,12 @@
 import gql from 'graphql-tag';
 
 const allPathsQuery = gql`
-  query pathsAndPowers {
```
lily-pytel: My bad, bad copy-pasting

scripts/import-cards.sh
```diff
@@ -7,4 +7,8 @@
 
 # This file also serves as an example of how one could import cards
 # to the production or staging environments.
+
+
 psql -h localhost -p 5433 -U postgres -d postgres -f import-cards.sql
+#psql -h localhost -p 5433 -U postgres -d postgres -f import-paths.sql
```
lily-pytel: Commented out?

next/components/all-paths.js
```diff
@@ -0,0 +1,28 @@
...
+import CardList from './card-list.js';
+import PropTypes from 'prop-types';
+
+import allPathsQuery from '../lib/queries/paths-query.js';
+
+export default function AllPaths(props) {
+  const { onPathClick } = props;
+  const { loading, error, data } = useQuery(allPathsQuery);
+
+  if (error) return <ErrorMessage message={error} />;
```
lily-pytel: This cries when there's an actual error because `message` is supposed to be a string. I think you want `message={error.message}`

next/components/all-paths.js
```diff
@@ -0,0 +1,28 @@
...
+  if (loading) return <div>Loading Paths...</div>;
+  if (!data || !data.paths) return null;
+
+  const paths = data.paths.nodes.filter(p => p.name !== 'No path selected');
+  return (
+    <CardList
+      onPathClick={onPathClick}
```
lily-pytel: This isn't in the props of `CardList` and doesn't seem to be used in that component but things seem to work? What dark magic is this???

next/components/all-powers.js
```diff
@@ -0,0 +1,28 @@
...
+  const { loading, error, data } = useQuery(allPowersQuery);
+
+  if (error) return <ErrorMessage message={error} />;
+  if (loading) return <div>Loading Powers...</div>;
+  if (!data || !data.powers) return null;
+
+  const powers = data.powers.nodes.filter(p => p.name !== 'No power selected');
+  return (
+    <CardList
+      onPathClick={onPowerClick}
```
lily-pytel: Same as the previous question. Also is this supposed to be `onPowerClick`?

next/components/header.js
```diff
@@ -17,16 +19,13 @@ class Header extends React.Component {
...
-        `}</style>
-        <style jsx>{`
-          .header {
             padding-top: 10px;
             max-height: 100px;
             overflow: hidden;
             position: fixed;
             font-weight: bold;
             max-width: 940px;
+            z-index: 2;
```
lily-pytel: 😅

scripts/import-cards.sh
alexmiddeleer: Yeah, it's more of an example than anything else.

next/components/all-paths.js
alexmiddeleer: > Looks good. Had a couple of clarification questions that may very well be answered in the PR to add path and powers to the deck being built

Yeah those are both fixed in the next PR

PR #71, opened 2019-09-07
Can add paths and powers to deck

alexmiddeleer:


next/components/all-paths.js
```diff
@@ -16,7 +16,7 @@ export default function AllPaths(props) {
   const paths = data.paths.nodes.filter(p => p.name !== 'No path selected');
   return (
     <CardList
-      onPathClick={onPathClick}
+      onCardClick={onPathClick}
```
lily-pytel: Ah. Never mind my questions from the other PR then

next/components/import-deck.js
```diff
@@ -52,8 +52,8 @@ export default function ImportDeck({
     data: powerData
   } = useQuery(allPowersQuery);
 
-  if (cardsError || pathError || powerError)
-    return <ErrorMessage message={error} />;
+  const error = cardsError || pathError || powerError;
+  if (error) return <ErrorMessage message={error} />;
```
lily-pytel: I believe this needs `message={error.message}`. Whenever I get errors it complains that it's expecting a string there

alexmiddeleer: All of these are likely broken. I will make a card to fix them everywhere.

PR #74, opened 2019-09-07
Add clear filters to the decks list page

lily-pytel:
This clears the filters BUT it doesn't work well on the card search one, since it seems like it handles its own state. The card ids are cleared but your selection and the suggestions are still there... I think it will require some more refactoring to basically make it not be the master of its own state so I figured I'll make another PR for that.

### DIFF
--- a/next/components/deck-search-form.js
+++ b/next/components/deck-search-form.js
@@ -30,6 +30,17 @@ export default function DeckSearchForm(props) {
     });
   };
 
+  const handleClear = e => {
+    e && e.preventDefault();
+    setName('');
+    setCardIds([]);
+    setFactionNames([]);
+    setIsOnlyFactions(true);
+    setUpdatedTime('150');
+    setAuthorName('');
+    onSubmit({});
+  };
+
   const { error, data } = useQuery(allCardsQuery);
 
   let cardSearchElement = null;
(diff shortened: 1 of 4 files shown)


next/components/deck-search-form.js
```diff
@@ -30,6 +30,17 @@ export default function DeckSearchForm(props) {
     });
   };
 
+  const handleClear = e => {
+    e && e.preventDefault();
```
alexmiddeleer: Only add these if necessary... They can cause weird bugs if we add tracking code later.

next/components/deck-search-form.js
```diff
@@ -30,6 +30,17 @@ export default function DeckSearchForm(props) {
...
+    setName('');
+    setCardIds([]);
+    setFactionNames([]);
+    setIsOnlyFactions(true);
+    setUpdatedTime('150');
+    setAuthorName('');
+    onSubmit({});
+  };
```
alexmiddeleer: could you get all of this from the state hooks, either by sharing variables or some built in hooks reset option?

next/components/deck-search-form.js
lily-pytel: Okay, I'll do the alternative of setting the button type to `button`

next/components/deck-search-form.js
lily-pytel: Looks like there's no reset functionality for hooks. The alternative is to put all these in a single object and then create a reset function like in the deck builder - do you prefer that?

https://stackoverflow.com/questions/54895883/reset-to-initial-state-with-react-hooks

alexmiddeleer:
Don’t worry about either of my comments, it’s gtg.

PR #75, opened 2019-09-07
Homepage new decks - no essence or upvotes yet (pr1)

alexmiddeleer:


next/components/deck-preview.js
```diff
@@ -0,0 +1,100 @@
...
+        }
+
+        :global(.manaIndicator span),
+        :global(.upvoteIndicator span) {
+          font-size: 16px;
+        }
+      `}</style>
+
+      <div className="deckPreview">
+        {/* TODO - duplicate code in deck-list.js*/}
```
lily-pytel: It won't be quite duplicated once the archetype column is between factions and essence 🤷‍♀

lily-pytel: I think encapsulating the indicators in their own components is good enough protection against evil code duplication

alexmiddeleer:
> So many indicators, feels like we're back at iVantage. Too close to launch for some tests, though?

🙈

PR #76, opened 2019-09-08
Decklist essence

alexmiddeleer:


next/lib/deck-utils.js
```diff
@@ -27,29 +27,3 @@ export const addCardToDeck = (deck, card) => {
 
   return newDeck;
 };
-
-export const collectDecksFactionsAndMana = decks => {
```
lily-pytel: Nice! TBH would not have been my choice because I hate the database (legibility and debugging abilities) and thought putting this effort on the user's browser was not a big deal, but I can see the logic of getting it away from the client. Either way good precedence for views.

alexmiddeleer: I *think* this is better but time will tell.

PR #77, opened 2019-09-08
Clear card selection filters

lily-pytel:
To clear card suggestions, I believe `DeckSearchForm` needs to have more control over `CardSearch`, which until this PR was quite independent. There might be a better way to do this but this seems like pretty basic React... maybe? I'm gonna be bummed and interested to hear if there's a better solution to reset a child component.

ANYWAY, there is a "bug" where if you enter something into the "includes cards" textbox and get a suggestion, you need to click on the "clear" button twice in order to clear the selection. This is because you need to click "out" of the suggestion first and then you can click on the button. Almost like there's a mask over the whole page you have to click out of. Seems like a `react-autosuggest` clownish behavior to me.

Looks like this now:

image

And all the filters can be cleared. It just doesn't work well if there's a suggestion in-flight. Ah well.


next/components/deck-search-form.js
```diff
@@ -9,34 +9,44 @@ import allCardsQuery from '../lib/queries/all-cards-query';
 import DeckSearchFormText from './deck-search-form-text';
 import DeckSearchFormUpdated from './deck-search-form-updated';
 
+const resetFilters = () => {
```
lily-pytel: I refactored the state into one big object because I thought I was getting into some async state update loop or something and that's why clearing while having a suggestion doesn't work. By the time I realized the issue... it was too late to refactor because life is short and release is close.

PR #80, opened 2019-09-08
Deck votes

alexmiddeleer:


alexmiddeleer:
> Looks good. Do we have the ability to upvote decks in v1?

I am not sure TBH. Might have been stupid to add it now. But since I was already in that area of the code and updating the view was fresh in my mind, I guess it made sense.

PR #82, opened 2019-09-08
Productionize (PR 1) - waiting on deck search rewrite to merge

alexmiddeleer:
Some basic stuff postgraphile recommends for your express server. The random query change was to prevent peeps from requesting 123123123 tournaments. That will be relevant with query whitelist PR that's next.

### DIFF
--- a/express/index.js
+++ b/express/index.js
@@ -2,8 +2,11 @@ const express = require('express');
 const { postgraphile } = require('postgraphile');
 const PgSimplifyInflectorPlugin = require('@graphile-contrib/pg-simplify-inflector');
 const ConnectionFilterPlugin = require('postgraphile-plugin-connection-filter');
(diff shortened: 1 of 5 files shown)


express/index.js
```diff
@@ -2,9 +2,14 @@ const express = require('express');
 const { postgraphile } = require('postgraphile');
 const PgSimplifyInflectorPlugin = require('@graphile-contrib/pg-simplify-inflector');
 const ConnectionFilterPlugin = require('postgraphile-plugin-connection-filter');
+const compression = require('compression');
+const helmet = require('helmet');
 
 const app = express();
 
+app.use(compression());
+app.use(helmet());
```
jtrussell: Nice. I wonder if we should add helmet to the next app too. I think it already uses compression under the hood.

PR #83, opened 2019-09-08
Query whitelisting  - waiting on deck search rewrite to merge

alexmiddeleer:
Note that currently the deck search is not whitelist-able, so it's broken right now. I will start working on that next. 

Obviously this is not ideal since we have to keep the whitelist file in sync with our next.js stuff, but it's a start. LMK if you have any better ideas for doing this. There are some automated tools, but we'd need to start using .graphql files and that seems like a big can of worms to open (need to move all of our queries into .graphql files and teach webpack how to handle them, and probably some additional stuff - I gave up at that point).

### DIFF
--- a/express/index.js
+++ b/express/index.js
@@ -4,7 +4,11 @@ const PgSimplifyInflectorPlugin = require('@graphile-contrib/pg-simplify-inflect
 const ConnectionFilterPlugin = require('postgraphile-plugin-connection-filter');
 const compression = require('compression');
 const helmet = require('helmet');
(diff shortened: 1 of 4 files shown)


jtrussell:
Interesting... thanks for diggging into this!

jtrussell:
Possible solution: put them in a node module.

alexmiddeleer:
> Possible solution: put them in a node module.

I think this is the way forward. Will make a story for it.

PR #86, opened 2019-09-09
Deck builder refactor part4

lily-pytel:
Last PR for the night. That slider is copied clean off the internet - don't judge.

Deck builder page with "view cards" selected:

image

Deck builder page with "view filters" selected:

image

Slides used in the decks page too:

image

I tell you, this really messes with Cypress' head. You have to tell it to go back and forth constantly.


e2e/cypress/page-objects/all.js
```diff
@@ -32,5 +32,9 @@ export default {
 
   // decks page
   deckFactionsPicker: '[data-cy="deckFactionsCell"] img',
-  deckEssencePicker: '[data-cy="deckEssenceCell"]'
+  deckEssencePicker: '[data-cy="deckEssenceCell"]',
+
+  // slider
+  leftLabel: '[data-cy="sliderLeftLabel"]',
+  rightLabel: '[data-cy="sliderRightLabel"]'
```
alexmiddeleer: I suggest `leftLabel` => `leftSlider` or make a `moveSliderLeft` fn. Could nix the comment as well.

next/pages/deck-builder.js
```diff
@@ -97,38 +44,38 @@ function DeckBuilderPage() {
...
+              setCardManaCosts={setCardManaCosts}
+              setSupertypes={setSupertypes}
+              setCardRarities={setCardRarities}
+            />
           )}
-          {currentTab === 'Powers' && (
-            <div className="collection" data-cy="deckBuilderPaths">
-              <AllPowers onPowerClick={onPowerClick}></AllPowers>
-            </div>
+          {!viewFilters && (
```
alexmiddeleer: I think this way if doing if-else is slightly more readable but /shrug (from react docs)
image

e2e/cypress/page-objects/all.js
lily-pytel: Changed

next/pages/deck-builder.js
lily-pytel: You're totally right. It was late.

e2e/cypress/page-objects/all.js
lily-pytel: Although I like the comments on this page

PR #87, opened 2019-09-09
fix: Allow users to save decks

evsheffield:
Super quick fix to the issue that Alex pointed out of not being able to save decks. TIL in Postgres the sequences associated with auto-increment IDs are permissioned separately from the tables themselves. 🤷‍♂ For reference: https://stackoverflow.com/questions/9325017/error-permission-denied-for-sequence-cities-id-seq-using-postgres


alexmiddeleer:
thanks! 🙇

PR #89, opened 2019-09-09
Card detail page

evsheffield:
Nothing too fancy happening here, just setting up the card detail page to match the design at https://mythgard-hub.github.io/card_detail_template.html


lily-pytel:
Suuuuuuuuuuuuper nitpicky but the image is supposed to be on the left according to the mock

image

next/components/card.js
```diff
@@ -1,15 +1,179 @@
...
+                <img
+                  src={RARITY_IMAGES[card.rarity.toLowerCase()]}
+                  className="rarity-icon"
+                />
+              </div>
+            </li>
+            <li className="card-detail">
+              <div className="card-detail-label">Subtype</div>
+              <hr />
+              <div className="card-detail-text">{card.subtype}</div>
```
lily-pytel: I think this needs that `firstLetterUppercase`

next/pages/card.js
```diff
@@ -1,35 +1,25 @@
...
-      rules
-    }
-  }
-`;
+export default withRouter(({ router }) => {
+  const { loading, error, data } = useQuery(cardQuery, {
+    variables: { id: parseInt(router.query.id, 10) }
+  });
+  if (loading) return null;
+  if (error) return <ErrorMessage message={`Error: ${error}`} />;
```
lily-pytel: Should be `message={error.message}`

next/components/gem-cost.js
```diff
@@ -0,0 +1,48 @@
...
+function GemCost({ costString }) {
+  return (
+    <>
+      <style jsx>{`
+        img {
+          height: 22px;
+          margin-right: 5px;
+        }
+      `}</style>
+      {costString
```
lily-pytel: I would put some safeguard here to make sure `costString` exists... you know me and my paranoia

next/pages/card.js
jtrussell: IIRC the `ErrorMessage` component also prepends `'Error: '` to the output as well.

next/components/card.js
```diff
@@ -1,15 +1,179 @@
...
+            </li>
+            {(card.atk || card.def) && (
+              <li className="card-detail">
+                <div className="card-detail-label">Attack/Health</div>
+                <hr />
+                <div className="card-detail-text">
+                  {card.atk}/{card.def}
```
jtrussell: LOL my brain was like "divide by zero?!".... haha.

next/components/gem-cost.js
jtrussell: What would be the appropriate response if `costString` doesn't exist? It's marked as a required prop but I suppose in prod that might not suffice since prop-type checks get disabled in prod (IIRC). Maybe a default prop value and an "unknown" gem cost symbol?

next/pages/card.js
lily-pytel: Right, hence only needing `message={error.message}`

next/components/gem-cost.js
lily-pytel: I saw always return da null or `[unknown]`

next/components/card.js
lily-pytel: I feel like the haha was followed by tears

lily-pytel:
@evsheffield sorry about that! I merged the PR fixing the error message, that's the only thing

next/components/card.js
evsheffield: I left this one intentionally unaltered because subtype can be pretty much anything and might consist of multiple words. `firstLetterUppercase` was mostly necessary for the enum types that don't have an immediately display friendly value stored in the database.

next/pages/card.js
evsheffield: 👍 I left this part unchanged from the original but didn't realize the error component has changed along the way. Thanks for catching that!

next/components/gem-cost.js
evsheffield: Assuming the card database is up to date there appear to be a few cards with no gem cost, so I'll default to `null`.

next/components/card.js
```diff
@@ -1,15 +1,179 @@
 import PropTypes from 'prop-types';
+import GemCost from './gem-cost.js';
+import { imagePathMedium as getImagePath } from '../lib/card.js';
+import { RARITY_IMAGES } from '../constants/rarities.js';
+import { SUPERTYPE_IMAGES } from '../constants/supertypes.js';
+
+const firstLetterUppercase = str => {
+  return str.charAt(0).toUpperCase() + str.slice(1).toLowerCase();
```
alexmiddeleer: could potentially move this to lib/card.js, maybe later though.

next/lib/queries/card-detail-query.js
```diff
@@ -0,0 +1,27 @@
+import gql from 'graphql-tag';
+
+export const cardQuery = gql`
+  query card($id: Int!) {
+    card(id: $id) {
```
alexmiddeleer: Soon every new query will need to go into the whitelist file... which I haven't merged yet. Just an fyi!

PR #90, opened 2019-09-10
Keep "session" alive while user is active

jtrussell:
This change does two things:

- Updates the JWT cookie expiration time whenever the user makes a request
- Clears the JWT cookie if it detects it is not longer valid (i.e. expired)

Since user info lives in the state of `_app`, there's also an interval to check to make sure that cookie still exists. Once the cookie gets eaten, the app will show you as being logged out.

Side note:

A best practice would be to instead maintain two separate JWTs. One with a long expiration time (maybe 7 days) and another with short fuse (e.g. 15 minutes). The longer living token grants you permission to request a short-lived token. The short-lived token would then give you access to app APIs. The rationale behind this strategy is that you get to revoke user access (within a 15-minute lag) while still not constantly hitting your DB or the auth provider with every request. This lets you do fancy things like sign folks out when you delete their account, otherwise they might be able to keep their session alive indefinitely... perhaps a fun consideration for down the road 😅.

### DIFF
--- a/next/pages/_app.js
+++ b/next/pages/_app.js
@@ -31,7 +31,8 @@ class MyApp extends App {
        * bundles. Thre should be a "browser" section in our package.json.
        */
       const fetch = require('isomorphic-unfetch');
-      const { JWT_COOKIE_NAME, SSR_HOST } = process.env;
+      const JWT_COOKIE_NAME = process.env.JWT_COOKIE_NAME;
+      const SSR_HOST = process.env.SSR_HOST;
       const token = ctx.req.cookies[JWT_COOKIE_NAME];
       const resp = await fetch(`${SSR_HOST}/auth/user/${token}`);
       user = await resp.json();
@@ -60,6 +61,20 @@ class MyApp extends App {
   updateUser = user => {
     this.setState({ user });
   };
+
+  componentDidMount() {
+    this.userCheckInterval = setInterval(() => {
+      if (!this.state.user) return;
+      const r = new RegExp(`\\b${process.env.JWT_COOKIE_NAME}\\b`);
+      if (!r.test(document.cookie)) {
+        this.setState({ user: null });
+      }
+    }, 5000);
+  }
+
+  componentWillUnmount() {
+    clearInterval(this.userCheckInterval);
+  }
 
   render() {
     const { Component, pageProps, apolloClient } = this.props;
--- a/next/pages/_app.js
+++ b/next/pages/_app.js
@@ -63,6 +63,8 @@ class MyApp extends App {
   };
 
   componentDidMount() {
+    // Don't bother with the interval if we're not signed in
+    if (!this.state.user) return;
     this.userCheckInterval = setInterval(() => {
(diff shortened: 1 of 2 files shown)


next/server-routes/auth.js
```diff
@@ -99,6 +104,30 @@ router.get(
   }
 );
 
+// If already logged in, update the JWT expiration and keep the party going
```
alexmiddeleer: Does this generally require a full page refresh?

next/pages/_app.js
```diff
@@ -61,6 +62,22 @@ class MyApp extends App {
...
+      if (!r.test(document.cookie)) {
+        this.setState({ user: null });
+      }
+    }, 5000);
+  }
+
+  componentWillUnmount() {
+    clearInterval(this.userCheckInterval);
+  }
+
```
alexmiddeleer: Is any of this code vulnerable to cookie tampering or does it not matter since we won't let them do anything mutationy without a valid JWT anyway?

alexmiddeleer: Not that we necessarily need perfect security at this stage... just wondering.

next/server-routes/auth.js
jtrussell: Nope - you'll get the benefit just for pinging the server. No full refresh required.

next/pages/_app.js
jtrussell: It's 100% not perfect. What's being done here is purely cosmetic though. The BE has ultimate control over creation/destruction of the actual JWT, and you can't extend your session without decrypting the cookie. You *could* give yourself some other random cookie the includes a string matching our JWT cookie name... if you did that while you were logged in the site wouldn't show you as being logged out until you did another SSR. However, for all intents and purposes you would be logged out - not API access, no valid JWT, no user.

next/server-routes/auth.js
alexmiddeleer: Is there a standard ping? Just wondering since next seems pretty static -  if the ssr is working, it's pretty much done after the initial handoff.

jtrussell: Good question. Pretty much all traffic is an opportunity to refresh the cookie. For example:

- Thanks to Next's code splitting, you'll make requests for additional bundles as you navigate from page to page. Of course, this is only good for the first time you go to a new page within the span of full page loads. I.e. it'll load the `/pages/cards.js` bundle only once between page loads.
- Any time you request data from the GraphQL server. This is a big one. This works because the next app proxies traffic to the graphql server and our `auth` middleware sits first in line before graphql (or the next engine, proper) get a chance to touch the request.

At least that's my understanding 😅.

jtrussell: Oh, also, I think `_app.js` makes a call to `/auth/user` on every page nav :D. We should probably maybe make that non-blocking if you already have a valid user.

alexmiddeleer: Oh right, I forgot next is a proxy for graphql (which I did, haha). I originally thought we were going to factor that out in production but that may never happen :P

PR #91, opened 2019-09-10
Productionize deck search (PR 1)

alexmiddeleer:
Not very friendly, but it seems to work and will allow us to create a whitelistable deck search query.


postgres/init.sql
```diff
@@ -221,6 +221,87 @@ CREATE TRIGGER update_deck_modtime
...
+    AND (authorName is NULL or to_tsvector('simple', account.username) @@ to_tsquery('simple',authorName || ':*'))
+    -- modification date filter
+    AND (deckModified is NULL or deck.modified >= deckModified)
+    -- cards filter
+    AND (numCards is null or numCards = 0 or (card.id in (card1, card2, card3, card4, card5)))
+
+    intersect
+
+    -- factions filter
+    SELECT deck.* from mythgard.deck left join mythgard.card_deck on (deck.id = card_deck.deck_id) left join mythgard.card on (card.id = card_deck.card_id) left join mythgard.card_faction on (card.id = card_faction.card_id) left join mythgard.faction on (faction.id = card_faction.faction_id)
```
evsheffield: The factions filter is the only part that makes me a little uneasy tbh, but I can't think of any alternatives right now and I certainly can't argue with results. We might at least be able to cut down on some of the duplication of this big select using a common table expression https://www.postgresql.org/docs/9.1/queries-with.html

alexmiddeleer: yeah it's definitely not ideal. I am not super comfortable with SQL yet. I'm thinking we might want to hold off on refactoring/optimizing. till later though.

jtrussell: A temporary (or permanent) view that abstracts away the joins might be an easy win for maintainability and also make life easier for the query plan optimizer.

alexmiddeleer: @jtrussell I definitely like that idea. I will try it.

jtrussell: This might be a dumb question (I'll be the first to admin a preference for working on the front-end), but could the additional intersects blocks be instead `AND (factionN is null or faction.id = factionN)`

alexmiddeleer: @jtrussell I don't think so, since after the left join there is still never a single row where faction.id = 1 and faction.id = 2. Unless I'm misunderstanding you.
The result of 
```
SELECT deck.id, deck.name as deckName, faction.id as factionId 
from mythgard.deck 
left join mythgard.card_deck on (deck.id = card_deck.deck_id) 
left join mythgard.card on (card.id = card_deck.card_id) 
left join mythgard.card_faction on (card.id = card_faction.card_id) 
left join mythgard.faction on (faction.id = card_faction.faction_id)
 ```
Is 
image

jtrussell: Ah, that makes sense - thanks! That also clears up why we have the group by. I wasn't thinking about the join right.

PR #93, opened 2019-09-12
Patch notes I suppose

alexmiddeleer:
Bit of a placeholder but gets the home page bit out of the way (though how we decide to wire some of the content up remains a question).

### DIFF
--- a/next/pages/index.js
+++ b/next/pages/index.js
@@ -2,6 +2,7 @@ import Layout from '../components/layout';
 import PageBanner from '../components/page-banner';
 import NewDecks from '../components/new-decks.js';
 import { useContext } from 'react';
+import Link from 'next/link';
 import { ThemeContext } from '../components/theme-context.js';
 
 const index = () => {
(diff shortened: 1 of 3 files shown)


next/pages/index.js
```diff
@@ -38,15 +39,24 @@ const index = () => {
...
+          text-decoration: none;
+        }
+        .patchNotes :global(.page-banner) {
+          height: 134px;
+        }
+      `}</style>
+      <div className="patchNotes">
+        <Link href="/patchNotes">
+          <a>
+            <PageBanner image={PageBanner.IMG_PATCH_NOTES}>
```
jtrussell: Just a heads up that this puts a block level element (`h1`) inside an an inline element (`a`).

alexmiddeleer: > Just a heads up that this puts a block level element (`h1`) inside an an inline element (`a`).

Isn't that ok in HTML5? Or do you think it's a bad thing?

jtrussell: Ah you're right, `a` is a special case they can hold block level elements. TIL.

PR #94, opened 2019-09-12
Footer I suppose

lily-pytel:
It's a little hacky but does anyone really care about one of the most ignored parts of any website? It's fine on pages with lots of content and on short pages it is a bit longer than the screen - hence the hack of making the height 97%. We can of course make it sticky and whatever but I am personally opposed - again, lots of real estate for something that has so little useful information. Let me know if you object.

Here it is on an empty page and a standard laptop screen: 

image

As you see, you can either see the footer or the banner - objections? I can play with the css some more though.

### DIFF
--- a/next/components/footer.js
+++ b/next/components/footer.js
@@ -1,0 +1,30 @@
+import { useContext } from 'react';
+import { ThemeContext } from './theme-context';
+
+// TODO - this is actually an essence indicator
+export default function Footer() {
+  const theme = useContext(ThemeContext);
+
+  return (
+    <div className="footer">
+      <style jsx>{`
+        .footer {
+          background-color: ${theme.footerBackgroundColor};
+          font-size: 11px;
+          text-align: center;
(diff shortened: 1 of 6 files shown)


next/components/footer.js
```diff
@@ -0,0 +1,30 @@
...
+        }
+
+        a {
+          text-decoration: none;
+          color: ${theme.footerLinkColor};
+        }
+      `}</style>
+      &copy; {new Date().getFullYear()} Mythgard |{' '}
+      <a href={`mailto:${process.env.EMAIL_MG_SUPPORT}`}>Contact</a> |{' '}
+      <a href="/privacy-policy">Privacy Policy</a>
```
jtrussell: `Link`?

lily-pytel: I was wondering why it wasn't a link on the banner? Is it because it's fake? Anyway no problem, fixed

PR #96, opened 2019-09-14
Add cookie max-age/expiration

jtrussell:
This let's our cookies die when they'd no longer be valid anyway.
Previously we had session cookies with a JWT expiration baked in. This
mean you could end up with a JWT... a valid cookie with an expired
JWT... if say you left your browser open for longer than the JWT
lifespan.


jtrussell:
Also, marking the cookies as HTTPS online in production.

PR #97, opened 2019-09-14
Fix prop type warnings and style tweak

jtrussell:
The page-banner component was marked as wanting only `string` children.
It now allows anything.

I made a little change to the footer styles while I was fixing the
prop-type warnings on the links there too.

### DIFF
--- a/next/components/footer.js
+++ b/next/components/footer.js
@@ -23,14 +23,17 @@ export default function Footer() {
           text-decoration: none;
           color: ${theme.footerLinkColor};
         }
+
+        * + *:before {
+          content: '|';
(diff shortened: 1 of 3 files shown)


next/components/footer.js
```diff
@@ -23,14 +23,17 @@ export default function Footer() {
           text-decoration: none;
           color: ${theme.footerLinkColor};
         }
+
+        * + *:before {
+          content: '|';
```
jtrussell: I guess it could have been `content: ' | ';` rather than using margin to match exactly 🤔.

lily-pytel: Potato | Potahto

PR #100, opened 2019-09-16
Preserve decks built while user is not logged in

jtrussell:
Basically this:

- Serializes your deck-in-progress every time you make an edit and
tosses that value into session storage
- Whenever you load the deck builder page, we first check session
storage to see if anything is there. If something is there, it gets
loaded into the deck builder
- Logged out users now see a "log in to save your deck" type button
instead of the usual save when on the deck builder page.
- The BE will now attempt to return you to the same page you were on
after you log in.

This will require updates after we layer in deck editing capabilities.
E.g. we may want to ignore edits in session storage if the deck ID for
those edits doesn't match the deck you're trying to edit now.

I added sessions support to the express server that wraps our Next app.
This allows us to flash the redirect URL and get it back after bouncing
off the auth provider. If feels a little heavy handed to add sessions
just for this... I'm open to other suggestions. Other common solutions
include flashing to a cookie or DB storage - neither of those felt
particularly appealing either.

### DIFF
--- a/e2e/cypress/integration/deck-builder-page.js
+++ b/e2e/cypress/integration/deck-builder-page.js
@@ -16,7 +16,11 @@ import {
 
 describe('Deck builder page', () => {
   beforeEach(() => {
-    cy.visit('/deck-builder');
+    cy.visit('/deck-builder', {
+      onBeforeLoad: win => {
+        win.sessionStorage.clear();
+      }
+    });
   });
   it('should have a happy path', function() {
     cy.get('[data-cy="header"]').should('be.visible');
@@ -123,15 +127,6 @@ describe('Deck builder page', () => {
       })
       .then(length => {
         expect(numCardsBeforeFilter).to.equal(length);
-
-        // finally try to save the deck and fail
-        cy.get('[data-cy="deckTitle"]').type('Floop the Pig');
-        cy.get('[data-cy="saveDeck"]').click();
-        cy.on('window:alert', str => {
-          expect(str).to.equal(
-            'Only logged in users can save deck. Please export your work, log in and come back.'
-          );
-        });
       });
   });
 
--- a/next/components/save-deck.js
+++ b/next/components/save-deck.js
@@ -2,6 +2,7 @@ import { useContext } from 'react';
 import PropTypes from 'prop-types';
 import { ApolloConsumer } from 'react-apollo';
 import Router from 'next/router';
+import Link from 'next/link';
 
 import UserContext from '../components/user-context';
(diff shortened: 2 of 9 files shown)


jtrussell:
Also, PR #100 WOOO 🎉

e2e/cypress/integration/deck-builder-page.js
```diff
@@ -16,7 +16,11 @@ import {
 
 describe('Deck builder page', () => {
   beforeEach(() => {
-    cy.visit('/deck-builder');
+    cy.visit('/deck-builder', {
+      onBeforeLoad: win => {
+        win.sessionStorage.clear();
```
alexmiddeleer: 🆒

next/pages/deck-builder.js
```diff
@@ -31,7 +31,35 @@ function DeckBuilderPage() {
...
+  // local/session storage, we need to make sure the code that loads/unloads a
+  // previously worked on decks is not run during an SSR.
+  useEffect(() => {
+    try {
+      const d = JSON.parse(sessionStorage.getItem('deckInProgress'));
+      if (!d) return;
+      const deckProperties = Object.keys(initializeDeckBuilder());
+      const isValidDeck = deckProperties.reduce((isValid, key) => {
+        return isValid && d.hasOwnProperty(key);
+      });
```
alexmiddeleer: I suppose we *could* try to populate whatever fields are matched rather than binary load deck or fail to load deck. But this is fine for an mvp. And this idea might just lead to problems.

next/components/save-deck.js
```diff
@@ -45,13 +46,18 @@ export default function SaveDeck({ deckInProgress }) {
...
-          margin-bottom: 10px;
-        }
-      `}</style>
+  // If user is not logged in, they must do so before saving
+  let saveButton;
+  if (!user || !user.id) {
+    saveButton = (
+      <Link href="/auth/google">
+        <a className="button" data-cy="saveDeck">
+          Login to Save
```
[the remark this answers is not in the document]
jtrussell: A fine question. I don't have super strong opinions either way, maybe disable the button and have a note on the page that they need to login first and their work will be saved 🤔.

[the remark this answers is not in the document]
jtrussell: I'll leave it as is for now, but my feelings won't be hurt if it gets changed.

next/pages/deck-builder.js
jtrussell: Half measures make me a little nervous. As my old robots professor used to say - "Do the right thing some of the time, never do the wrong thing."

PR #102, opened 2019-09-16
Integrate new deck query

alexmiddeleer:


express/query-checker.js
```diff
@@ -0,0 +1,20 @@
+const allowedQueries = require('./whitelisted-queries.js');
+
+//prettier-ignore
+// eslint-disable-next-line
+const jhashcode=s=>{for(var i=0,h;i<s.length;i++)h=Math.imul(31,h)+s.charCodeAt(i)|0;return h};
```
jtrussell: I this typically how hashes are gotten in the gql world? Since this runs on the BE we have access to legit hashing functions.

alexmiddeleer: Nope, I was just prototyping with stackoverflow answers. I updated it to use sha1. Good call.

PR #105, opened 2019-09-17
Add margin images

jtrussell:


jtrussell:
OK for round 1?

Screen Shot 2019-09-16 at 10 06 48 PM

cc @PapaCapricorn

PapaCapricorn:
Short answer: Looks good, don't let this slow you down.

Long-winded answer: I think what we are going for is a layout that feels dynamic when you scroll or re-size, without creating any weird or unexpected results. This is difficult to evaluate with a static image, but I trust that whatever you produce will work for round 1.

ps - You are just asking for a review of the margin image and not any of the other page elements, correct?

jtrussell:
Cool, short term we may want to just shrink those images down a little bit. Long term, and after you get to play with it, we can figure out if we're meeting the desired experience.

re: PS - correct. Though feel free to reach out on discord regarding other things.

PR #107, opened 2019-09-17
Homepage top decks

alexmiddeleer:


next/lib/deck-queries.js
```diff
@@ -238,4 +238,20 @@ export const deckPreviewToDeck = d => {
   };
 };
 
+export const topDeckPreviewsQuery = gql`
+  query deckPreview {
+    deckPreviews(orderBy: DECK_CREATED_DESC, first: 3, filter: {
```
evsheffield: Perhaps a small point depending on how we intend to use this feature, but something that just occurred to me. Do we want this to pull the top decks by date the *deck* was created or by the date that somebody flagged it as a *featured deck*, or for that matter some arbitrary ordering that we designate? e.g. Suppose there are 6 "featured decks" but the oldest one is really the most bau5 so we want to float that one to the top.

alexmiddeleer: Yep, that's definitely an issue to be addressed later. It's not actually extremely easy to sort these in another way right now. This was extremely easy, so I'm going to punt and call this MVP.

PR #111, opened 2019-09-18
Add some article boilerplate

jtrussell:
Just prepping for when we have article content to include.

Screen Shot 2019-09-17 at 10 46 31 PM

Screen Shot 2019-09-17 at 10 46 40 PM


next/components/article.js
```diff
@@ -0,0 +1,70 @@
+import PropTypes from 'prop-types';
+import { useContext } from 'react';
+import { ThemeContext } from './theme-context';
+
+function Article({ title, url, description, date }) {
+  const theme = useContext(ThemeContext);
+
+  const now = new Date();
```
evsheffield: We should be able to leverage the existing time util for this: https://github.com/mythgard-hub/mythgard-hub/blob/801ef8a57e078151e2b8ec4960484e52fca7b699/next/lib/time.js

next/components/article.js
```diff
@@ -0,0 +1,70 @@
...
+      </div>
+    </>
+  );
+}
+
+Article.propTypes = {
+  title: PropTypes.string.isRequired,
+  url: PropTypes.string.isRequired,
+  description: PropTypes.string.isRequired,
+  date: PropTypes.instanceOf(Date)
```
evsheffield: Should this one also be required? Seems like there may be sad times with `date.getTime()` if this prop isn't provided.

next/components/article.js
jtrussell: Oh nice, I didn't know this existed 😅. Looks very familiar ;) https://github.com/mythgard-hub/mythgard-hub/pull/68/files#diff-55725476055ab916956f3bdf9b2cd93aR8-R26

next/components/article.js
jtrussell: Yup, sure should.

PR #113, opened 2019-09-18
Fix power path export

evsheffield:
I went to fix some unit tests and found a couple of small bugs with export which should now be fixed.

- Power/path were showing up as "Power: undefined" in exports instead of being empty. This one the unit test caught for us!
- Power/path were always showing up as empty when exporting decks from the deck detail page


evsheffield:
@lily-pytel Nice! 👍 In the interest of preserving the git history as a separate commit I think we can just merge one of them first and resolve conflicts in the other.

PR #115, opened 2019-09-19
Make card import script production ready.

alexmiddeleer:
Most important thing - card import now upserts "on id". Thus we can import the card database over and over without breaking anyone's deck. 

I also added the other fields we support, the most significant being faction.

### DIFF
--- a/scripts/import-cards.sql
+++ b/scripts/import-cards.sql
@@ -29,7 +29,7 @@ CREATE TEMPORARY TABLE t (
 -- We use -1 to denote variable attack and defense values.
 -- However, the csv uses X or * to denote this. A regex is
 -- used to convert these to -1.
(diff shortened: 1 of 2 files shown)


scripts/import-cards.sql
```diff
@@ -17,23 +17,51 @@ CREATE TEMPORARY TABLE t (
...
+    ,supertype = excluded.supertype
+    ,subtype = excluded.subtype
+    ,atk = excluded.atk
+    ,def = excluded.def
+    ,mana = excluded.mana
+    ,gem = excluded.gem
+    ,rarity = excluded.rarity
+;
+
+truncate table mythgard.card_faction;
```
alexmiddeleer: I think this is fine.. but lmk if you think of a reason it's not fine.

jtrussell: :burn: all the things 👍

PR #116, opened 2019-09-19
Fix e2e tests

jtrussell:
Mostly adding/update `data-cy` things. Also made individual cards in deck lists links, thanks to cypress for catching that one.


jtrussell:
(Still not expecting them to work on Travis)

PR #117, opened 2019-09-19
Swap back postgres creds for auth check

jtrussell:
Hey @evsheffield - I probably just had the wrong actual passwords in my secrets file but we should review these too just in case.

It looked like postgraphile was using the generic PGUSER/PGPASS creds and the auth checks were using the EXPRESS_* creds. I wasn't sure if there's a need for two separate creds or what the vision was for where we'd use which ones.

### DIFF
--- a/next/server-routes/auth.js
+++ b/next/server-routes/auth.js
@@ -6,9 +6,9 @@ const jwt = require('jsonwebtoken');
 const faker = require('faker');
 
 const client = new PgClient({
-  user: process.env.EXPRESS_PGUSER,
+  user: process.env.PGUSER,
   database: process.env.PGDATABASE,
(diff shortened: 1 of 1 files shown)


evsheffield:
It's probably not strictly necessary to have two different users, but I think it's nice to keep them separate for bookkeeping purposes and to be able to limit access to just what is needed. The generic PGUSER is the one that bootstraps the DB so it actually gets owner-level privileges on everything. The idea was that our custom express routes really only need to interact with the accounts table so we can restrict its access to just those relevant pieces of the DB. If that seems reasonable I would prefer to keep the separate credentials for this connection.

jtrussell:
Sounds good - just send along the correct passwords so I can update secrets in the deploy scripts 👍.

PR #118, opened 2019-09-19
Move logout link to the welcome banner

evsheffield:
Super small change to move the Logout link to the banner. I also ditched the little "down arrow" next to the user name because it felt misleading and with Logout in the banner there were too many arrows.

### DIFF
--- a/next/components/welcome-banner.js
+++ b/next/components/welcome-banner.js
@@ -77,9 +77,9 @@ export default function WelcomeBanner() {
               {user.username || 'Welcome'}
             </a>
           </Link>
-          <Link href="/auth/logout">
-            <a className="welcome-banner-link">Logout</a>
-          </Link>
+          <a href="/auth/logout" className="welcome-banner-link">
+            Logout
+          </a>
         </div>
       )}
     </div>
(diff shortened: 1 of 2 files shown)


next/components/welcome-banner.js
```diff
@@ -75,11 +71,16 @@ export default function WelcomeBanner() {
...
-            {user.username || 'Welcome'}
-          </a>
-        </Link>
+        <div>
+          <Link href="/account">
+            <a className="welcome-banner-link welcome-banner-link-username">
+              {user.username || 'Welcome'}
+            </a>
+          </Link>
+          <Link href="/auth/logout">
```
jtrussell: This one should just be a regular ol' anchor tag since there's no next page for it.

jtrussell: It'll still work with `Link`, we'll just try to load a JS bundle for a client side render before giving up and falling back on an SSR.

jtrussell: (and I know I'm the guy who put it initially with a `Link` and no inner `a` at all 🤣)

evsheffield: Hah I didn't even notice that, thanks! `Link` is just so pleasing to read...

PR #123, opened 2019-09-22
Fix deck links from profile and SomeCards

jtrussell:
Also, added a loading message to the deck page so you don't see an error
message while the query is in flight.


alexmiddeleer:
A failing e2e test would be preferable but this is fine

jtrussell:
I owe you two, one for the profile and one for the search.

PR #126, opened 2019-09-22
Fix temple snake and other spawns with no rarity

alexmiddeleer:


scripts/optimize-card-images.sh
```diff
@@ -22,6 +22,7 @@
 
 # First copy the full sized images to the "m" (medium) folder. Medium is currently the largest
 # size we will use on the web.
+mkdir core/m
```
alexmiddeleer: just some random cruft sneakin in again :P

PR #129, opened 2019-09-22
Cards page 2 column layout

alexmiddeleer:
image

image
image

### DIFF
--- a/e2e/cypress/integration/cards-page.js
+++ b/e2e/cypress/integration/cards-page.js
@@ -138,6 +138,28 @@ describe('Cards Page', function() {
       })
       .then(length => {
         expect(allCardsLength).to.equal(length);
+
+        // basic test - clear filters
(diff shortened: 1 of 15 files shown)


e2e/cypress/integration/cards-page.js
```diff
@@ -138,6 +138,28 @@ describe('Cards Page', function() {
       })
       .then(length => {
         expect(allCardsLength).to.equal(length);
+
+        // basic test - clear filters
+        cy.get(`${rarityPicker} img:first`).click();
+        cy.get(`${superTypePicker} img:first`).click();
```
jtrussell: Love to see these not care about `img`... or `:first` for that matter. What do you think of just `data-cy='rarity-common-toggle'` or some such?

alexmiddeleer: good idea. done. first is still necessary otherwise we need to make bigger changes.

PR #132, opened 2019-09-24
New player guide

evsheffield:
The styles are a little funky as some of the styles have introduced global changes or styling that is inconsistent with our theme so far (e.g. links as orange instead of white). I tried to limit the global changes that were taking place to largely non-disruptive ones while respecting Mark's new styles and making sure they we would have what we need for the privacy page and other articles. There's been some chatter about these styles elsewhere (https://trello.com/c/NK5gvMxs/7-update-privacy-policy) so a follow up may be in order.

### DIFF
--- a/next/pages/_app.js
+++ b/next/pages/_app.js
@@ -87,6 +87,10 @@ class MyApp extends App {
           <HooksApolloProvider client={apolloClient}>
             <Head>
               <link rel="icon" type="image/x-icon" href="/static/favicon.ico" />
+              <link
+                href="https://fonts.googleapis.com/css?family=Vollkorn:400,400i,600,600i,700,700i&display=swap"
+                rel="stylesheet"
+              ></link>
               <meta charSet="utf-8" />
               <meta
                 name="viewport"
--- a/next/pages/_app.js
+++ b/next/pages/_app.js
@@ -88,7 +88,7 @@ class MyApp extends App {
             <Head>
               <link rel="icon" type="image/x-icon" href="/static/favicon.ico" />
               <link
-                href="https://fonts.googleapis.com/css?family=Vollkorn:400,400i,600,600i,700,700i&display=swap"
+                href="https://fonts.googleapis.com/css?family=Vollkorn:400,400i&display=swap"
                 rel="stylesheet"
               ></link>
               <meta charSet="utf-8" />
--- a/next/pages/index.js
+++ b/next/pages/index.js
@@ -32,7 +32,7 @@ const index = () => {
           border-bottom: ${theme.border};
         }
 
-        .npgBanner {
+        .newPlayerGuideBanner {
           clear: both;
           margin-top: 15px;
           float: left;
@@ -41,13 +41,13 @@ const index = () => {
           width: 874px;
         }
 
-        .npgBannerImage {
+        .newPlayerGuideBannerImage {
           z-index: -1;
           float: left;
           margin-bottom: 1px;
         }
 
-        .npgBannerText {
+        .newPlayerGuideBannerText {
           color: ${theme.fontColorHeading};
           font-size: 2em;
           font-weight: 700;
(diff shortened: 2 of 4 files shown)


next/pages/_app.js
```diff
@@ -87,6 +87,10 @@ class MyApp extends App {
           <HooksApolloProvider client={apolloClient}>
             <Head>
               <link rel="icon" type="image/x-icon" href="/static/favicon.ico" />
+              <link
```
jtrussell: I'm making a mental note to sync up with @PapaCapricorn on necessary this font is. Maybe there's a built in font that would suffice? I haven't tested the page weight delta yet, but this feels like a lot to add. At the very least, do we need each of the weights? I could only find where `400` was being used.

alexmiddeleer: At least it's a google font... also, I'm ignorant about this, do we need a backup font? I suppose their cdn might die at some point or get DOS'd

next/pages/index.js
```diff
@@ -32,6 +32,39 @@ const index = () => {
           border-bottom: ${theme.border};
         }
 
+        .npgBanner {
```
alexmiddeleer: I'm not a big fan of acronyms. I know it's less typing, but I'm a pretty forgetful dude and I get grumpy two months later when I'm like wtf is npg... same with `. mb ` below. Not sure at first blush what that is. Mythgard banner?

alexmiddeleer: Also, I feel like my colleagues love to hide bad names behind acronyms sometimes (not you guys, peeps at work). Should we just call it `mg-article` instead of `.npg`? That seems somewhat more generic and reusable as far as names go. (I'm okay with one or two acronyms for prefixing generic words with the site name, like `mg-box`. Ok I'm done ranting now. This is fine to merge even without my acronym insecurities.

next/pages/_app.js
evsheffield: Good call, I kind of just copied it blindly from the mockups but at the very least we can restrict the weights. TBH I'm not super worried about the availability of their CDN. If it ever goes away permanently for some reason we can deal with it then. 🤷‍♂

next/pages/index.js
evsheffield: Completely fair. Some of these were copied direct from the mockups so I'm not sure exactly what they are (mythgard banner for `mb` was my best guess) but I also perpetuated the style. `npg` is new player guide where I was too lazy to write the full thing, but I'll update it for the sake of clarity.

PapaCapricorn:
Awesome, thanks Evan! I’m looking forward to seeing the results when this goes live on the dev site. 

I see some mentions here so I’ll try to briefly respond, and please just say the word if I can further clarify any of this….

**Acronyms:** I am very guilty of using some bad and unclear acronyms (and other component names) on the mockups and for this I apologize. This is due to a combination of ignorance and attempts to take short cuts. Believe it or not, I have been going through and trying to clear things up bit by bit (although probably not quickly enough to be useful). 

TBH the process for applying/translating the styles defined in the templates is still a bit mysterious to me, ie: how much is lifted directly by copy/pasting code vs. eyeballing and then re-building things? So I’m not sure how much of an impact my bad names have when you are referring to the mockups. Ultimately I expect this part of the process will continue to improve, and please do not hesitate to ask if things are ever unclear - I am easy to reach and usually quick to respond.

**Fonts:**  I am leaning towards defaulting back to Exo 2 for body copy (<p> tag) based on some feedback, and will coordinate with Evan on how to accomplish this, but it is nothing critical atm.

Having said that, I’d rather not shut the door on adding one or two more Google fonts in the future. I appreciate any and all efforts to reduce load times, and will spend some time educating myself about fonts and page weight. But it would surprise me if using more than one font for a site is backbreaking. Is a speed test tool such as https://tools.pingdom.com/ useful for evaluating this? 

**Funky styles and global changes:** A follow up on the topic of global styles would be great. I don’t want this to slow things down, but as time allows it would be useful to have a list of the observed inconsistencies or disruptive styles. Without a better understanding of what is going on under the hood, it is hard for me to know where the conflicts are. 

In some cases what you are seeing may be on purpose - for example, there _is_ more than one color for links on the mock-ups (depending on context) and that is intentional. In other cases it is probably me making a hack job of it. I am looking forward to re-organizing things in a more efficient way based on input from all of you when the time is right.

PR #147, opened 2019-09-27
Add Impel to Articles page

PapaCapricorn:
Implementing this request from Noah: 

`If we could also make a late inclusion to the articles page, another source has started producing content and diversifying that page a bit more would be splendid.
This article should be added to the articles page: https://teamdgn.net/2019/09/27/the-many-uses-of-impel/`

### DIFF
--- a/next/components/article-list.js
+++ b/next/components/article-list.js
@@ -39,6 +39,14 @@ function ArticleList({ max }) {
       description:
         'Control decks are in it for the long haul.  Traditionally, they play a patient, defensive game...',
       date: new Date('2019-09-22T17:24:18.280Z')
+    },
+    {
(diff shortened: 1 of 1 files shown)


PapaCapricorn:
@lily-pytel I would tend to agree with you about the re-writes, but to clarify: this summary was just copied directly from the article itself, it did not originate with me or MGH. We might choose to use a shorter or longer excerpt, but it would feel awkward to me to re-write someone else's copy unless it was something they asked for.

lily-pytel:
@PapaCapricorn ah, I see. Never mind then!

lily-pytel:
@alexmiddeleer any reason not to merge this?

PapaCapricorn:
Merge away plz :)

PR #148, opened 2019-09-28
Add RLS for decks

jtrussell:
Add row-level security for decks.

The reason I wasn't seeing RLS having any effect locally was that my local secrets were configured to authenticate postgraphile using the `postgres` (admin) user as opposed to the `postgraphile` (non-admin) user. Another reason for multiple users and perhaps an entirely separate one to create the initial schema.

### DIFF
--- a/postgres/init.sql
+++ b/postgres/init.sql
@@ -119,6 +119,20 @@ INSERT INTO mythgard.card_deck("deck_id", "card_id", "quantity") VALUES (2, 1, 1
 INSERT INTO mythgard.card_deck("deck_id", "card_id", "quantity") VALUES (3, 1, 1), (3, 2, 1), (3, 3, 1), (3, 4, 1), (3, 5, 1), (3, 6, 1);
 INSERT INTO mythgard.card_deck("deck_id", "card_id", "quantity") VALUES (4, 1, 1), (4, 2, 1);
 
+ALTER TABLE mythgard.deck ENABLE ROW LEVEL SECURITY;
+CREATE POLICY deck_admin_all ON mythgard.deck TO admin USING (true) WITH CHECK (true);
+CREATE POLICY deck_all_view ON mythgard.deck FOR SELECT USING (true);
+CREATE POLICY deck_all_create ON mythgard.deck FOR INSERT WITH CHECK (true);
+CREATE POLICY deckupdate_if_author
+  ON mythgard.deck
+  FOR UPDATE
+  USING ("author_id" = mythgard.current_user_id())
+  WITH CHECK ("author_id" = mythgard.current_user_id());
+CREATE POLICY deckdelete_if_author
+  ON mythgard.deck
+  FOR DELETE
+  USING ("author_id" = mythgard.current_user_id());
+
 CREATE TABLE mythgard.deck_vote (
   id SERIAL PRIMARY KEY,
   deck_id integer,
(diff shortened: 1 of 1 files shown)


postgres/init.sql
```diff
@@ -2,6 +2,11 @@ DROP SCHEMA IF EXISTS mythgard CASCADE;
 
 CREATE SCHEMA mythgard;
 
+CREATE OR REPLACE FUNCTION mythgard.current_user_id()
```
jtrussell: Moved this up since it needed to be defined before it was used in the deck table policies.

postgres/init.sql
```diff
@@ -101,6 +106,25 @@ INSERT INTO mythgard.deck("name", "path_id", "power_id", "author_id") VALUES ('c
...
+CREATE POLICY deck_all_view ON mythgard.deck FOR SELECT USING (true);
+-- Non-admins can create
+CREATE POLICY deck_all_create ON mythgard.deck FOR INSERT WITH CHECK (true);
+-- Rows can only be updated by their author
+CREATE POLICY deckupdate_if_author
+  ON mythgard.deck
+  FOR UPDATE
+  USING ("author_id" = mythgard.current_user_id())
+  WITH CHECK ("author_id" = mythgard.current_user_id());
+-- Rows can only be updated by their author
```
alexmiddeleer: perhaps you meant `-- Rows can only be deleted by their author`

lily-pytel:
I'm happy to test it. Could I get some instructions?  Sounds like a good way to get into the swing of things.

alexmiddeleer:
@lily-pytel I suggest putting a debugger in right before the graphql mutation fires off and update the id to something you shouldn't be allowed to update. Alternatively you could copy the graphql request "as curl" and then modify the variables. Burp or Postman would be the GUI options (there might be more too)

jtrussell:
Getting migrations rolling is a solid idea, especially since we basically will never run `init.sql` in prod again.

I'm going to pass on creating a migration at this point. *How* we want to handle migrations is probably worth a team discussion, and implementing a non-trivial migration strategy (i.e. works with docker-compose + prod, doesn't require redundant logic, etc.) probably deserves its own PR. If we want to leave this open until that's in place, that's cool, I/we can add the appropriate migration then.

alexmiddeleer:
Nah, thats ok. As long as I can copy paste this into a psql prompt and not
blow up production I’m fine with it.

lily-pytel:
@jtrussell Is this a valid way to test this?

image

It did let me delete deck 1 ("dragons"), of which I am not the author

jtrussell:
@lily-pytel yep, that should be a fine way to test it. Can you double check that the pguser and express_pguser settings in your .env file match the template?

lily-pytel:
@jtrussell oh ya, I bet it is not

postgres/init.sql
```diff
@@ -101,6 +106,25 @@ INSERT INTO mythgard.deck("name", "path_id", "power_id", "author_id") VALUES ('c
 INSERT INTO mythgard.deck("name", "modified") VALUES ('all_factions', '2019-05-1 00:00:00');
 INSERT INTO mythgard.deck("name", "modified") VALUES ('norden aztlan', '2019-01-1 00:00:00');
 
+ALTER TABLE mythgard.deck ENABLE ROW LEVEL SECURITY;
+-- Admin users can make any changes and read all rows
+CREATE POLICY deck_admin_all ON mythgard.deck TO admin USING (true) WITH CHECK (true);
+-- Non-admins can read all rows
+CREATE POLICY deck_all_view ON mythgard.deck FOR SELECT USING (true);
+-- Non-admins can create
+CREATE POLICY deck_all_create ON mythgard.deck FOR INSERT WITH CHECK (true);
```
evsheffield: Perhaps less important from a security perspective but it might also be reasonable to have `WITH CHECK ("author_id" = mythgard.current_user_id())` on inserts also so that users can only create decks belonging to themselves.

jtrussell: Great idea 👍 - added.

PR #153, opened 2019-09-29
Sometimes cards appeared in 2 columns on deckbuilder page

alexmiddeleer:
Would a windows user mind testing this? I can't repro on mac but I was fiddling on Friday on my windows pc and it seemed to fix this.


lily-pytel:
@alexmiddeleer what was causing the original issue to happen? I can test

alexmiddeleer:
Basically the right side panel was too wide and caused the flex box to
collapse a bit. It seemed to be caused by this text box.

PR #154, opened 2019-10-01
Limit card quantities by rarity

lily-pytel:
- trying to import too many cards results in importing the max by rarity
- trying to click on a card that is already at max will do... nothin'

### DIFF
--- a/next/constants/rarities.js
+++ b/next/constants/rarities.js
@@ -7,6 +7,13 @@ export const RARITY_IMAGES = {
   mythic: `${cdn}Filter-Icons_0000s_0000s_0003_M.png`
 };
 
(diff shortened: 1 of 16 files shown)


next/constants/rarities.js
```diff
@@ -7,6 +7,13 @@ export const RARITY_IMAGES = {
   mythic: `${cdn}Filter-Icons_0000s_0000s_0003_M.png`
 };
 
+export const RARITY_MAX_CARDS = {
```
lily-pytel: do you prefer me to put this in the db like `mythgard.essence_costs` instead?

alexmiddeleer: @lily-pytel that's only in there because we need it for a sql fn.

PR #155, opened 2019-10-01
Fix some tests

alexmiddeleer:
Going to see if I can fix Travis.

### DIFF
--- a/e2e/cypress/integration/cards-page.js
+++ b/e2e/cypress/integration/cards-page.js
@@ -157,6 +157,7 @@ describe('Cards Page', function() {
           'setup for clear button test'
         );
         cy.get('[data-cy="cardSearchClear"]').click();
+        cy.wait(500);
         return getPagingTotalAsInt();
       })
       .then(length => {
--- a/e2e/cypress/integration/cards-page.js
+++ b/e2e/cypress/integration/cards-page.js
@@ -33,28 +33,31 @@ describe('Cards Page', function() {
         return getPagingTotalAsInt();
       })
       .then(length => {
-        expect(length).to.be.below(allCardsLength);
+        expect(length).to.be.below(allCardsLength, 'search text filter works');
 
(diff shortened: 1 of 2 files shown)


postgres/init.sql
```diff
@@ -308,19 +308,19 @@ create function mythgard.search_decks(deckName varchar(255), authorName varchar(
     intersect
 
     -- cards filter
-    SELECT deck.* from mythgard.deck left join mythgard.card_deck on (deck.id = card_deck.deck_id) left join mythgard.card on (card.id = card_deck.card_id) left join mythgard.card_faction on (card.id = card_faction.card_id) left join mythgard.faction on (faction.id = card_faction.faction_id)
+    SELECT deck.* from mythgard.deck left join mythgard.card_deck on (deck.id = card_deck.deck_id) left join mythgard.card on (card.id = card_deck.card_id) left join mythgard.card_faction on (card.id = card_faction.card_id)
```
alexmiddeleer: these extra joins were just cruft

e2e/cypress/integration/cards-page.js
```diff
@@ -146,12 +152,16 @@ describe('Cards Page', function() {
         return getPagingTotalAsInt();
       })
       .then(length => {
-        expect(allCardsLength).to.be.above(length);
+        expect(allCardsLength).to.be.above(
+          length,
+          'setup for clear button test'
+        );
         cy.get('[data-cy="cardSearchClear"]').click();
+        cy.wait(500);
```
alexmiddeleer: I can't think of anything else to do here, sadly. I don't think a network request should happen b/c of the cache, so can't tell cy to wait for that. There's probably a better thing to do here as cypress is pretty clever but I'm going to have to research that later. LMK if you can think of anything.

lily-pytel:
Haven't seen that green checkmark in a thousand years

alexmiddeleer:
I know it. it feels... good!

PR #156, opened 2019-10-01
Author articles, 1st attempt

PapaCapricorn:
A first attempt at adding Authors to the list on the Articles page, with image attached for formatting reference...tell me where I went wrong and I'll try to fix it. Cheers!

image


jtrussell:
Nice! The only issue is using a `div` for the author will cause line breaks that I'm not sure were intended:

Screen Shot 2019-10-01 at 8 22 05 PM

To get around that, you can either use a `span` instead 👍.

jtrussell:
One of thing to be aware of - a quirk of using JSX is whitespace between elements gets clobbered. In regular ol' HTML, your whitespace will collapse nicely into a breaking space. Not so in JSX.

This is why there didn't end up being a space between the end of the title and the word `by`. Working around the quirk is a little annoying, usually folks insert a `{' '}` for a breaking space or `&nbsp;` for a non-breaking space.

PapaCapricorn:
Thanks @jtrussell - made some revisions based on your suggestions.

This version also includes a new file for the incoming 7 Things article which can be ignored for now.

alexmiddeleer:
@PapaCapricorn We can't really stage files this way. Mind deleting those changes and moving them to a different branch? Also, before you work more on that page we should discuss how to host articles on MGH. In the short run I'm fine to do a link to the reddit post. We could even do it the way you did it with a page in next/pages. Generally however, next.js pages are generic like "article" and then you load *data* into them (such as article content). You wouldn't make a next.js page for a specific article.

PapaCapricorn:
Thanks for explaining that @alexmiddeleer - the article page has been removed pending further discussion.

alexmiddeleer:
@PapaCapricorn I added a few small updates. LMK if it's ready to merge.

PR #160, opened 2019-10-03
Essence factions to deck

lily-pytel:
Looks like this:

image

### DIFF
--- a/express/whitelisted-queries.js
+++ b/express/whitelisted-queries.js
@@ -295,6 +295,9 @@ module.exports = [
       path {
         name
       }
+      deckPreviews {
+        ${deckPreviewsFragment}
+      }
     }
   }
 `,
--- a/next/components/deck.js
+++ b/next/components/deck.js
@@ -6,6 +6,8 @@ import DeckDelete from './deck-delete';
 import { initializeDeckBuilder } from '../lib/deck-utils';
 import { deckCardsQuery } from '../lib/deck-queries';
 import DeckCardsTable from './deck-card-table';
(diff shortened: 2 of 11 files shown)


express/whitelisted-queries.js
```diff
@@ -295,6 +295,9 @@ module.exports = [
       path {
         name
       }
+      deckPreviews {
```
alexmiddeleer: Might rename this thing at some point or do a big cleanup - I designed this kinda clumsily. This code is okay for now, although you might have been able to do two separate requests (1 network call thanks to batching) to avoid touching a query.

lily-pytel:
@alexmiddeleer 

> Might suggest adding a single cypress assertion (just the "some essence cost exists" sort), but that's optional.

You're right... that was so lazy of me to not add a little assertion.

Also I made some global style changes. Let me know what you think.

next/components/deck.js
```diff
@@ -53,12 +70,39 @@ export default function Deck({ deck }) {
...
         }
 
         .deck-author {
           margin-bottom: 20px;
         }
 
+        .deck-stats {
+          width: 100%;
+          margin-left: 20px;
+          width: fit-content;
```
alexmiddeleer: I have never seen this before! Looks like it might not be allowed by Edge. https://caniuse.com/#search=fit-content

lily-pytel: Oops, also I have width there twice - fixed. It's in `CardSelectionItem` too. I can't believe Edge doesn't support it!!! It's the best!

PR #161, opened 2019-10-03
Minor formatting details on Home page and Events page; graded horizontal rules

PapaCapricorn:
**SUMMARY**

Attempting to make a few minor style changes to the Home page and the Events page:

* Added columnHeader class to layout.js
* Replaced < h2 > with columnHeader on index.js
* Added hr.orangeGrade class to layout.js
* Added hr.orangeGrade elements to index.js

_Intended formatting for reference (image):_

image

* Also: Minor adjustments to spacing and alignment on Events page

**LONG-WINDED, DURDLY RAMBLINGS**

* We currently have a few different examples of graded horizontal rules on the site (outlined below). 

* These HRs are handled slightly differently from each other, and the new addition (hr.orangeGrade) combines definitions from different approaches.

* At some point we might want to make these elements consistent, but it is not a priority by any means. I can work to make this happen, but would need some guidance.

* For now I believe the changes submitted in this PR will do the trick.

**GRADED HORIZONTAL RULES**

* **hr.bgrad** (defined in layout.js):
https://github.com/mythgard-hub/mythgard-hub/blob/master/next/components/layout.js#L148  _Defined using RGBa and theme colors_

  appears on [Articles](https://www.mythgardhub.com/articles) page:
  https://github.com/mythgard-hub/mythgard-hub/blob/master/next/components/article.js#L37

* **hr.ograd** (defined and appears on [Account](https://www.mythgardhub.com/account) page):
https://github.com/mythgard-hub/mythgard-hub/blob/master/next/pages/account.js#L80
_Defined using hex colors_

* **hr** (Defined in layout.js - appears on [card-search-filters.js](https://www.mythgardhub.com/cards) and others):
https://github.com/mythgard-hub/mythgard-hub/blob/master/next/components/layout.js#L143
_Defined: how?_

* **NEW: hr.orangeGrade** (Defined in layout.js - appears on index.js)
https://github.com/mythgard-hub/mythgard-hub/blob/minor-formatting-details/next/components/layout.js#L168
_Defined using RGBa and hex colors_


alexmiddeleer:
@PapaCapricorn I'm impressed with how close you got, though there were a few issues which you can see above. It's probably impossible to try to get it perfect every time without actually running the site on your machine to see what your changes did. 
image
I'm guessing this gradient was supposed to fade out on both sides, not just the left?

alexmiddeleer:
Also, I needed to keep those elements as H2 or else they lost the other styling they needed. But I removed the double border.

PapaCapricorn:
> there were a few issues which you can see above.

I'm seeing the '/' which you added here: what are the rules for adding a '< />?

> I'm guessing this gradient was supposed to fade out on both sides, not just the left?

Actually the different fades were intentional, with double-sided being reserved for centered header elements (like the column headers), and the single-sided for left or right justified labels with text or other elements directly above and below. But I'm open to re-visiting if there are other opinions.

IMO this one is ready to go. What is the next step?

alexmiddeleer:
> I'm seeing the '/' which you added here: what are the rules for adding a '< />?

You always need a `/` before the closing caret on singleton tags.

PapaCapricorn:
> You always need a `/` before the closing caret on singleton tags.

Ok thanks. Is anything else required here before merging?

alexmiddeleer:
I just need to do a final round of testing but it should be good to go

PR #162, opened 2019-10-03
7 things article

PapaCapricorn:
- This probably needs lots of edits to make it work, but the basic formatting for the article body is there.

- Once we have this in place with a URL I will add it to the Articles (Media?) list page...


alexmiddeleer:
@PapaCapricorn Looks good. Please submit a PR with a link on the articles page when you're ready.

PR #165, opened 2019-10-06
Deck builder - disable save for invalid deck sizes

lily-pytel:
- No e2e test for this because we haven't figured out how to test authentication so we don't test saving at all

- Changed the color of disabled buttons because it's just really light. Looks like it's only used for buttons


next/lib/deck-utils.js
```diff
@@ -37,3 +37,14 @@ export const getQuantity = (card, oldQuantity, newQuantity) => {
 
   return Math.min(desiredQuantity, max);
 };
+
+export const getCardCount = deck => {
+  try {
```
alexmiddeleer: Might want to throw a debug statement in the catch

lily-pytel:
@alexmiddeleer So the save button was not disabled when you only had 2 cards? I didn't put anything to prevent saving, only UI controls to disable the save. I can change it to happen on the mutation\database end, though

lily-pytel:
> Awesome, thanks. Constants for `50` and `200` would be ideal but not required.

@alexmiddeleer oh yeah, added

PR #173, opened 2019-10-12
Scroll to top deck

lily-pytel:
- I am not able to reproduce what is happening in production exactly
- This solution moves the scrolling for decks to happen after query complete
- Does require using @apollo/react-hooks... I know, yet another provider. We can slowly migrate the other ones to use this. I did it at work and it's the same, but I didn't want to muddy this PR even more
- For `AllDecks`, the initial load is ignored, but for `SomeDecks` it is not. If we ever put the search filters in the URL, we'll have to figure out what to do there
- There is a one second delay after noting the initial load because sometimes it looks like `onCompleted` was called twice and sometimes not... jeez. Making that delay seems to help.


next/lib/navigation-utils.js
```diff
@@ -0,0 +1,7 @@
+export const scrollToTopOfList = listRef => {
```
alexmiddeleer: I might just move this to a ui-utils library... probably not going to have tons more navigation-utils. Also, could rename this since it doesn't seem list specific.

next/components/all-decks.js
```diff
@@ -1,19 +1,31 @@
...
       offset: currentPage * pageSize
+    },
+    onCompleted: _ => {
+      if (!dataLoaded.current) {
+        setTimeout(() => {
+          dataLoaded.current = true;
+        }, 1000);
+      } else {
+        scrollToTopOfList(listRef);
+      }
```
alexmiddeleer: Could you instead use a `useEffect` callback that watches `data ` and only does something if `data` is populated? If not, I'm okay to stick with this solution for now.

lily-pytel: Since useEffect also gets called during the initial load as well as subsequent loads, we'd still have to use a ref to keep track of whether this is the first load or the next ones. It would be basically the same... arguably more confusing

next/lib/navigation-utils.js
lily-pytel: Fair enough

PR #175, opened 2019-10-12
Events mvp

alexmiddeleer:
Mind making sure it works on your machine Lily?


lily-pytel:
Definitely! I started looking last night, will continue shortly

next/components/large-table.js
```diff
@@ -0,0 +1,41 @@
+import { useContext } from 'react';
+import PropTypes from 'prop-types';
+import { ThemeContext } from './theme-context';
+
+export default function LargeTable(props) {
```
lily-pytel: Nice!

next/components/layout.js
```diff
@@ -311,6 +325,16 @@ function Layout({ title, desc, children }) {
             margin-bottom: 25px;
           }
 
+          .external-link::after {
+            background-image: url('https://cdn.mythgardhub.com/icons/newWindow.svg');
```
lily-pytel: I believe this symbol means that the link opens in another window, so I would make this an actual link that opens a new window next to the regular link, or make the link open a new window. Obviously nbd though.

alexmiddeleer: I'm actually anti-new window these days... there are some security and performance reasons. I think I will leave it as is for now unless others agree. I view the symbol more as "warning: this might take you to a different host". Reasons why I'm anti new window:
 https://developers.google.com/web/tools/lighthouse/audits/noopener

lily-pytel: That's totally fine with me - I never open new windows unless Isabelle demands it. I just wouldn't use that symbol in that case, since I believe it explicitly means opening a new window - at least that's what I expected when I click it. I also wish it turned the same blue when you hover over it... anyway, it's beyond not important

PR #181, opened 2019-10-17
Cards page - Preserve filters when returning to search feature from a different page

alexmiddeleer:


alexmiddeleer:
There's probably a cleaner approach... but it does work.

Issue #184, opened 2019-10-19
Card hover in deckbuilder is preventing fast deckbuilding


alexmiddeleer:
On pointer devices (e.g. desktop), the hover should be delayed slightly so that cards can be quickly added to a deck by clicking on them.

![image](https://user-images.githubusercontent.com/5400639/67135605-77bc1680-f1e9-11e9-9ed1-dbe7f7e2cd54.png)

Issue #183, opened 2019-10-19
On mobile, tapping on a card to add it to a deck in the deckbuilder causes unwanted zoom-in


alexmiddeleer:
There should not be a card hover on mobile

![image](https://user-images.githubusercontent.com/5400639/67135564-08462700-f1e9-11e9-8b07-5ab0824b843b.png)

PR #189, opened 2019-10-19
Card hover images: delay on desktop, remove on mobile

lily-pytel:
Delay showing the hover images, but only on devices that have hover

I didn't know about this media type... so much I don't know yet

Fiexes #184 and #183

### DIFF
--- a/next/components/card-list-item.js
+++ b/next/components/card-list-item.js
@@ -29,14 +29,23 @@ export default function CardListItem({ card, onClick, options }) {
           display: inline-block;
           position: relative;
         }
-        // hover image
-        .imgWrapper:hover::before {
+        // bigger version of the image (hidden until hover)
+        .imgWrapper::before {
           content: url(${imgPathMedium});
           width: ${hoverImageWidth}px;
           position: absolute;
           top: -${hoverImageVerticalOffset}px;
           left: -${(hoverImageWidth - smallImageWidth) / 2}px;
           z-index: 2;
+          opacity: 0;
+        }
+
+        @media (hover: hover) {
+          // Show the hover image (but only on devices that have hover)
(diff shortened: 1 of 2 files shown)


next/components/card-list-item.js
```diff
@@ -29,14 +29,24 @@ export default function CardListItem({ card, onClick, options }) {
...
-        .imgWrapper:hover::before {
+        // bigger version of the image (hidden until hover)
+        .imgWrapper::before {
           content: url(${imgPathMedium});
-          width: ${hoverImageWidth}px;
           position: absolute;
           top: -${hoverImageVerticalOffset}px;
           left: -${(hoverImageWidth - smallImageWidth) / 2}px;
           z-index: 2;
+          visibility: hidden;
```
lily-pytel: This makes sure the image actually doesn't "exist" while not hovering. Before it was there so when you hovered over path or power, you got the enlarged image of the card below it and couldn't really click. I've never used this visibility business - cool! It's like display but it can be delayed because it's apparently a transition.

next/components/card-list-item.js
```diff
@@ -29,14 +29,24 @@ export default function CardListItem({ card, onClick, options }) {
           display: inline-block;
           position: relative;
         }
-        // hover image
-        .imgWrapper:hover::before {
+        // bigger version of the image (hidden until hover)
+        .imgWrapper::before {
           content: url(${imgPathMedium});
-          width: ${hoverImageWidth}px;
```
lily-pytel: It looks like this is ignored :-(

lily-pytel:
@alexmiddeleer Added pointer and reduced the delay to 0.5 seconds. I struggled an embarrassing amount of time last night trying to come up with a "perfect" delay... it was conflicting.

Issue #182, opened 2019-10-19
Hovering over a card in deck builder with pointer device should show image


alexmiddeleer:
You should be able to use similar css hover code as we do on the /cards page.

lily-pytel:
Is it supposed to have a little thumbnail or is it when hovering over the name? Or the entire row?

PapaCapricorn:
> Is it supposed to have a little thumbnail or is it when hovering over the name? Or the entire row?

@lily-pytel For now I would say entire row. Thumbnails have been proposed, and they might co-exist (as they do on EWC). I'm not sure if you mean just including the card art, or are referring specifically to a smaller clickable hotspot area. Either way, in the short term just making the entire row is probably easiest, and would be my recommendation.

PR #190, opened 2019-10-20
Deck builder deck hover

lily-pytel:
* Pretty much the same as the card hover, except I put the small version here. Is that okay or do we want the full version?
* Also no delay here... seems unnecessary.
* I put the hover effect just on the name cell. It was a bit unseemly to do on the entire row.

Looks like this (but imagine a mouse on the name of the card):
image

@alexmiddeleer @PapaCapricorn

### DIFF
--- a/next/components/deck-card-table-row.js
+++ b/next/components/deck-card-table-row.js
@@ -53,7 +53,6 @@ export default function DeckCardsTableRow({ card, deleteCard, quantity }) {
           .imgWrapper:hover::before {
             visibility: visible;
             opacity: 1;
-            transition-delay: 0.7s;
           }
         }
       `}</style>
--- a/next/components/deck-card-table-row.js
+++ b/next/components/deck-card-table-row.js
@@ -10,14 +10,13 @@ export default function DeckCardsTableRow({ card, deleteCard, quantity }) {
   const backgroundColor = cardMainColor(card, theme);
   const color = backgroundColor ? theme.cardTableName : 'white';
   const imagePath = imagePathSmall(card.name, card.set || undefined);
-  const smallImageWidth = IMAGE_SIZES.smallImageWidthPortrait;
 
   return (
     <tr key={card.id} data-cy="deckCardRow">
       <style jsx>{`
         td {
           padding: 2px 5px 5px 5px;
-          border: 1px solid #03080a;
+          border: ${theme.cardTableBorder};
         }
         .deck-delete-card {
           cursor: pointer;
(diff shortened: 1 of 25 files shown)


next/components/deck-card-table-row.js
```diff
@@ -0,0 +1,93 @@
...
+  const color = backgroundColor ? theme.cardTableName : 'white';
+  const imagePath = imagePathSmall(card.name, card.set || undefined);
+  const smallImageWidth = IMAGE_SIZES.smallImageWidthPortrait;
+
+  return (
+    <tr key={card.id} data-cy="deckCardRow">
+      <style jsx>{`
+        td {
+          padding: 2px 5px 5px 5px;
+          border: 1px solid #03080a;
```
alexmiddeleer: Using this in 2 spots now. Time for a color variable? "border-light" perhaps? I know it's dark, but we have a dark theme so it's basically a light border.

next/components/deck-card-table-row.js
```diff
@@ -0,0 +1,93 @@
...
+  const backgroundColor = cardMainColor(card, theme);
+  const color = backgroundColor ? theme.cardTableName : 'white';
+  const imagePath = imagePathSmall(card.name, card.set || undefined);
+  const smallImageWidth = IMAGE_SIZES.smallImageWidthPortrait;
+
+  return (
+    <tr key={card.id} data-cy="deckCardRow">
+      <style jsx>{`
+        td {
+          padding: 2px 5px 5px 5px;
```
alexmiddeleer: could do `2px 5px 5px` Not even sure I prefer that notation, this is more of an FYI.

next/components/deck-card-table-row.js
```diff
@@ -0,0 +1,93 @@
...
+
+        @media (hover: hover) {
+          // Show the hover image (but only on devices that have hover)
+          .imgWrapper:hover::before {
+            visibility: visible;
+            opacity: 1;
+          }
+        }
+      `}</style>
+      <td>{card.mana < 0 ? 'X' : card.mana}</td>
```
alexmiddeleer: we might want to move this to a formatter function in card utils. It's used in 2 spots

next/components/deck-card-table-row.js
lily-pytel: I am not a fan of that shortened notation, personally

next/components/deck-card-table-row.js
lily-pytel: Done. Let me know if the variable is too specific. We already have other borders so I prefer to give them more specific names (for the theme-context, of course, not the color itself, which is just a super dark gray) @alexmiddeleer

next/components/deck-card-table-row.js
lily-pytel: Done

lily-pytel:
@alexmiddeleer good point, I changed it so the image appears when you hover everywhere on the name line

lily-pytel:
Fixes #182

PR #192, opened 2019-10-22
Search filters from url - decks page edition

alexmiddeleer:


alexmiddeleer:
@lily-pytel Last one for search filters in the url, I promise!

lily-pytel:
It might be what is causing the tests to fail as well

alexmiddeleer:
> Nice! Only thing is that we'll only show decks from the last 150 days now, right? That's intentional? My only concern is that there no way to know that it's what happening (there's no showing x out of y like PUMA has)

Well, I am just trying to match it to the form default. This is prod, but it's not actually the last 150 days - it's all of our decks.

image

lily-pytel:
@alexmiddeleer you're right, fair enough

PR #201, opened 2019-10-27
Import event results via csv

alexmiddeleer:


alexmiddeleer:
It worked on production 😆

PR #205, opened 2019-10-29
Deck voting

alexmiddeleer:


lily-pytel:
Oh man oh man, the most anticipated feature

next/components/deck-export.js
```diff
@@ -57,7 +57,7 @@ DeckExport.propTypes = {
       id: PropTypes.number,
       name: PropTypes.string
     }),
-    deckPath: PropTypes.shape({
+    deckPower: PropTypes.shape({
```
lily-pytel: This again?!

alexmiddeleer: I don't even know. This is mgh's weirdest issue.

lily-pytel:
image

Is there not supposed to be an upvote symbol there that takes you to the login page?

Also when I logged in and went to this page (http://localhost:3001/deck?id=3), I got this:

image

I know it's because that deck has no author and that case is impossible but I think it's worth putting a safeguard there anyway

Then I tried to go to the cats deck and vote but got:

image

Console:

image

Network was more helpful: 

image

In conclusion, I sadly can't vote yet! 😆

alexmiddeleer:
@lily-pytel Thanks for the review! I think that last error was just that you needed to restart the express container because of the query whitelist. But definitely some good catches.

alexmiddeleer:
@lily-pytel Also, mind sanity-checking the last commits or if you have time, trying to hack it?

lily-pytel:
@alexmiddeleer yep, looking now.

> I think that last error was just that you needed to restart the express container because of the query whitelist. But definitely some good catches.

That would be strange because I basically destroy all containers and rebuild between every branch or database change or PR or almost every day, really

PR #208, opened 2019-11-03
Unify head layout

Edswor:
Move all the head metas to the layout component, so instead of havinbf head component in _app, deck, card pages it's only in layout.


Edswor:
Yes. As I know more about react nextjs I'm trying to make the code cleaner.

Issue #207, opened 2019-11-02
Deck essence calculating incorrectly


alexmiddeleer:
Essence for cards with multiple factions are being calculated incorrectly.

https://www.mythgardhub.com/deck?id=714

The influence for waystone garden is doubled. Replace it with another random mythic and it calculates essence correctly as 18.2. With waystone, 20600 essence is reported.

PR #210, opened 2019-11-03
Fix essence cost multi faction

lily-pytel:
@alexmiddeleer what do you think? There's a lot of duplicate queries in there, I know. I am not sure there is a way to do this without separate queries, though.


lily-pytel:
@alexmiddeleer is that what a migration supposed to be?

alexmiddeleer:
> @alexmiddeleer is that what a migration supposed to be?

It's just something I can copy/paste into the production database before a deploy and it will work, basically.

lily-pytel:
Fixes #207

PR #211, opened 2019-11-03
Deck builder edit path power

lily-pytel:
PR 1 for the deck builder refactoring

- First step in the big deck builder refactor - adding an edit button in the path and power rows.
- This did mean that `tab-group` can no longer maintain its own state regarding which tab is selected... it's gotta come from above.
- `tab-group` is also using the context hook.
- The table needed to be a bit bigger to look good, so refactored the spaces a bit to look more like the template (the margins are more narrow there so there is space for 3 columns of cards). The rest of the pages look okay on desktop and the same on mobile.
- Events page: moved the loading and error to be inside the layout.
- There are no new tests on purpose, since the final stage might be quite different in the end.

This is the big change:
image

This seems like a lot of refactoring but it's only step 1. Sorry if it has too much!

### DIFF
--- a/next/lib/string-utils.js
+++ b/next/lib/string-utils.js
@@ -8,7 +8,7 @@ export const firstLetterUppercase = str => {
 
 export const ordinalized = i => {
   try {
-    if (!i) return '';
+    if (!i && i !=== 0) return '';
 
     const j = i % 10,
       k = i % 100;
--- a/next/lib/string-utils.js
+++ b/next/lib/string-utils.js
@@ -6,9 +6,17 @@ export const firstLetterUppercase = str => {
   }
 };
 
+function isInt(value) {
+  return (
+    !isNaN(value) &&
+    parseInt(Number(value)) == value &&
+    !isNaN(parseInt(value, 10))
+  );
+}
+
 export const ordinalized = i => {
   try {
-    if (!i && i !== 0) return '';
+    if (!isInt(i)) return '';
 
     const j = i % 10,
       k = i % 100;
--- a/next/lib/string-utils.js
+++ b/next/lib/string-utils.js
@@ -8,6 +8,8 @@ export const firstLetterUppercase = str => {
 
 export const ordinalized = i => {
   try {
+    if (!i) return '';
+
     const j = i % 10,
       k = i % 100;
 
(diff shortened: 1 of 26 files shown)


next/pages/event.js
```diff
@@ -10,106 +10,112 @@ import FactionsIndicator from '../components/factions-indicator.js';
 import EssenceIndicator from '../components/essence-indicator.js';
 import { dbDateToDisplayDate } from '../lib/time.js';
 import { tournamentWithResultsQuery as tourneyQuery } from '../lib/tournament-queries.js';
+import { ordinalized } from '../lib/string-utils';
 import Router from 'next/router';
 
-// Copied from stackoverflow, so untested.
```
alexmiddeleer: Thought this comment was unnecessary huh? :P

next/pages/event.js
```diff
@@ -10,106 +10,112 @@ import FactionsIndicator from '../components/factions-indicator.js';
...
   });
 
+  let pageTitle = 'Mythgard Hub | Event Results';
+  let tournamentDecksTable = null;
+
   if (error) {
-    return <ErrorMessage message={error.message} />;
+    tournamentDecksTable = <ErrorMessage message={error.message} />;
   } else if (loading) {
-    return <div>Loading...</div>;
```
alexmiddeleer: thanks for removing this, it was annoying me

next/pages/event.js
lily-pytel: Yeah, because I was going to add tests, and then I forgot 😨

lily-pytel: I'll add them in a sec after merging the other branch

lily-pytel:
@alexmiddeleer Added tests and all the other branches and some PR comments. Do you need to do another pass?

next/lib/string-utils.js
```diff
@@ -8,7 +8,7 @@ export const firstLetterUppercase = str => {
 
 export const ordinalized = i => {
   try {
-    if (!i) return '';
+    if (!i && i !=== 0) return '';
```
alexmiddeleer: Might just want to add an "Is Int" func.
```
function isInt(value) {
  return !isNaN(value) && 
         parseInt(Number(value)) == value && 
         !isNaN(parseInt(value, 10));
}
``` 
from stackoverflow.

lily-pytel: Nice. That did the trick

PR #213, opened 2019-11-06
Deck builder action buttons

lily-pytel:
View with cards

image

Import mode

image

Empty mode

image

### DIFF
--- a/next/components/deck-builder-action-buttons.js
+++ b/next/components/deck-builder-action-buttons.js
@@ -1,0 +1,50 @@
+import PropTypes from 'prop-types';
+import DeckExport from '../components/deck-export';
+import SaveDeck from '../components/save-deck';
+
+export default function DeckBuilderActionButtons(props) {
+  const {
+    cardCount,
+    importMode,
+    deckId,
+    deckInProgress,
+    setDeckInProgress,
+    onClear
+  } = props;
+
+  if (!cardCount || importMode) return null;
+
+  return (
+    <div className="action-buttons" data-cy="deckBuilderActions">
+      <style jsx>{`
+        .action-buttons {
+          display: flex;
+          justify-content: space-between;
+        }
+        :global(.save-deck-container),
+        :global(.deck-export-container),
+        .clear-button-container {
+          width: 106px;
+        }
+      `}</style>
+      <SaveDeck
+        deckId={deckId}
+        deckInProgress={deckInProgress}
+        setDeckInProgress={setDeckInProgress}
+      />
+      <div className="clear-button-container">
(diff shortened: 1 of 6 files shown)


next/components/deck-builder-action-buttons.js
```diff
@@ -0,0 +1,50 @@
...
+    <div className="action-buttons" data-cy="deckBuilderActions">
+      <style jsx>{`
+        .action-buttons {
+          display: flex;
+          justify-content: space-between;
+        }
+        :global(.save-deck-container),
+        :global(.deck-export-container),
+        .clear-button-container {
+          width: 106px;
```
alexmiddeleer: In this case it might be practical, but in general I prefer to avoid fixed widths and let things flow a bit with the browser window. Don't worry about it but keep it in mind for the future - Mark isn't going to scream if the button is sometimes 120px wide and sometimes 90.

next/components/import-deck.js
```diff
@@ -64,35 +82,44 @@ export default function ImportDeck({
...
         }
       `}</style>
+      <button
+        onClick={onClick}
+        data-cy="importDeckButton"
+        className="import-button"
+      >
+        Import Decklist
+      </button>
+      <button onClick={_ => setImportMode(false)} data-cy="cancelImportMode">
```
alexmiddeleer: is _ not a lint error?

next/components/deck-builder-action-buttons.js
lily-pytel: @alexmiddeleer I actually made this 31% and then remembered you didn't like it and changed to fixed width 😆 . Just can't win!

next/components/import-deck.js
lily-pytel: @alexmiddeleer no, neither e nor _ are lint errors in this case (I think it's because it's inline), but I've gotten used to doing _ since it denotes "I'm not gonna use this thing, but it's there"

next/components/deck-builder-action-buttons.js
alexmiddeleer: > @alexmiddeleer I actually made this 31% and then remembered you didn't like it and changed to fixed width 😆 . Just can't win!

I'm sorry. I'm a pain. I don't like either I'm afraid. I like my elements to grow to fit whatever space their in and let margin, padding, flex-grow and the width of the browser control everything (in my perfect world at least).

lily-pytel: Ah, that makes sense! I'll stick to that in the future

alexmiddeleer: With the exception of inline-elements, which I let act like inline-elements and not grow.

PR #215, opened 2019-11-07
Fix to essence cost multiplied by votes

lily-pytel:
It's basically the same as when it was being multiplied by factions. Joins... ugh. The view needs to be regenerated to be fixed so hopefully the migration takes care of it.

### DIFF
--- a/postgres/init.sql
+++ b/postgres/init.sql
@@ -363,7 +363,7 @@ BEGIN
 END;
 $$ language 'plpgsql';
 
-CREATE OR REPLACE FUNCTION deck_essence_cost (IN deckId INTEGER) 
+CREATE OR REPLACE FUNCTION mythgard.deck_essence_cost (IN deckId INTEGER) 
 RETURNS INTEGER AS $$
   DECLARE
     essence_cost INTEGER;
@@ -383,7 +383,7 @@ RETURNS INTEGER AS $$
   END;
   $$ language 'plpgsql';
 
-CREATE OR REPLACE FUNCTION deck_votes (IN deckId INTEGER) 
+CREATE OR REPLACE FUNCTION mythgard.deck_votes (IN deckId INTEGER) 
 RETURNS INTEGER AS $$
   DECLARE
     votes INTEGER;
@@ -404,8 +404,8 @@ CREATE OR REPLACE VIEW mythgard.deck_preview as
          deck.name as deck_name,
          deck.created as deck_created,
          array_agg(DISTINCT faction.name) as factions,
-         deck_essence_cost(deck.id)::int as essence_cost,
-         deck_votes(deck.id)::int as votes
+         mythgard.deck_essence_cost(deck.id)::int as essence_cost,
+         mythgard.deck_votes(deck.id)::int as votes
   FROM mythgard.deck
   JOIN mythgard.card_deck
     ON card_deck.deck_id = deck.id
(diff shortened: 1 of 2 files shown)


postgres/init.sql
```diff
@@ -380,7 +380,22 @@ RETURNS INTEGER AS $$
...
+    SELECT count(DISTINCT deck_vote)::int into votes
+    FROM mythgard.deck
+   LEFT JOIN mythgard.deck_vote
+     On deck_vote.deck_id = deck.id
+   WHERE deck.id = $1
+   GROUP BY deck.id;
+    
+    RETURN votes;
   END;
   $$ language 'plpgsql';
```
alexmiddeleer: Probably could just be `language sql`, and then you don't need the SELECT into stuff. See other `language sql` funcs.

postgres/init.sql
```diff
@@ -400,10 +415,6 @@ CREATE OR REPLACE VIEW mythgard.deck_preview as
     ON (card.id = card_faction.card_id and card_faction.faction_id is not null)
   LEFT JOIN mythgard.faction
     On faction.id = card_faction.faction_id
-  LEFT JOIN mythgard.essence_costs
-    On essence_costs.rarity = card.rarity
```
alexmiddeleer: Oops, I guess that join was basically dead code

postgres/init.sql
```diff
@@ -390,7 +405,7 @@ CREATE OR REPLACE VIEW mythgard.deck_preview as
          deck.created as deck_created,
          array_agg(DISTINCT faction.name) as factions,
          deck_essence_cost(deck.id)::int as essence_cost,
```
alexmiddeleer: Maybe we just needed `deck_essence_cost(distinct deck.id)::int as essence_cost,`? If that's even legal?

postgres/init.sql
```diff
@@ -380,7 +380,22 @@ RETURNS INTEGER AS $$
     GROUP BY deck.id;
     
     RETURN essence_cost;
-  
+  END;
+  $$ language 'plpgsql';
+
+CREATE OR REPLACE FUNCTION deck_votes (IN deckId INTEGER)
```
alexmiddeleer: should be mythgard.deck_votes probably

postgres/init.sql
```diff
@@ -390,7 +405,7 @@ CREATE OR REPLACE VIEW mythgard.deck_preview as
          deck.created as deck_created,
          array_agg(DISTINCT faction.name) as factions,
          deck_essence_cost(deck.id)::int as essence_cost,
-         count(DISTINCT deck_vote)::int as votes
+         deck_votes(deck.id)::int as votes
```
alexmiddeleer: mythgard.deck_votes

postgres/init.sql
lily-pytel: Why did we do `update_modified_column` in `plpgsql`? Also I kind of like it more - looks more like a normal function to me

postgres/init.sql
lily-pytel: Nooo it did not like that. I'm not that sure what that was supposed to do but the compiler or builder or whatever was not a fan

postgres/init.sql
alexmiddeleer: @lily-pytel That's cool. The answer to the first question is probably that I copy pasted it from something.

postgres/init.sql
alexmiddeleer: /shrug okies

PR #218, opened 2019-11-10
Deck sorting

alexmiddeleer:


next/components/deck-search-form.js
```diff
@@ -105,16 +119,30 @@ export default function DeckSearchForm(props) {
           margin: 10px 0;
           width: 100%;
         }
+        :global(.deck-search-form-sort) {
```
lily-pytel: Any reason not to have this in the component itself?

postgres/migrations/migration-1.4.1.sql
```diff
@@ -24,7 +24,7 @@ CREATE OR REPLACE VIEW mythgard.deck_preview as
          deck.name as deck_name,
          deck.created as deck_created,
          array_agg(DISTINCT faction.name) as factions,
-         deck_essence_cost(deck.id)::int as essence_cost,
+         mythgard.deck_essence_cost(deck.id)::int as essence_cost,
```
lily-pytel: This migration is useless really as it still has the votes bug. I would delete it and just use 1.4.2

next/components/deck-search-form.js
```diff
@@ -127,13 +155,17 @@ export default function DeckSearchForm(props) {
...
             flex-direction: column;
             margin-bottom: 20px;
           }
 
           .action-buttons input {
             margin-bottom: 15px;
+            margin-right: 0;
+          }
+
+          :global(.deck-search-form-sort) {
```
lily-pytel: Any reason not to have this in the component?

lily-pytel: Never mind, I see why!

next/components/deck-search-form.js
lily-pytel: I saw why, nvm!

lily-pytel: Actually never mind the never mind - looks like this global is only used here, so why not have it in the component?

postgres/migrations/migration-1.4.1.sql
alexmiddeleer: Oops!

alexmiddeleer:
> This is awesome! My only request is that if we sort by rating, that we show it somewhere - even somewhere temporary.

It's very tempting but I'll let the team voice interest first I think.

Issue #204, opened 2019-10-28
Add links to top level section pages from title banners


PapaCapricorn:
- Image and text in title banners on sub level pages should link to the top level section page as a secondary means of navigation

- For example, clicking the image or the word CARDS on the banner of the [Foul Harvest card page](https://www.mythgardhub.com/card?id=2) should link back to the [top level Cards section page](https://www.mythgardhub.com/cards) 

- See [image](https://user-images.githubusercontent.com/39966530/67645124-315b6b80-f8fd-11e9-8f49-147a2f836a1e.png) for reference

- See also related issue: [Display DECKS section banner on individual deck pages](https://github.com/mythgard-hub/mythgard-hub/issues/203)

![image](https://user-images.githubusercontent.com/39966530/67645124-315b6b80-f8fd-11e9-8f49-147a2f836a1e.png)

wajeehsheikh:
HI, 
I would like to help out with this issue.
Thanks,
Waj

PapaCapricorn:
Hi Wajeeh -

Thanks for your interest. Are you on the Mythgard Hub Discord server?

If not you can join here: https://discord.gg/RBX3zeG

Cheers,
PapaC

YSavir:
To clarify, is someone working on this issue already?

If not, I'd be glad to take on it. As someone new to Next, it's a great intro. I've actually already started on it just as a way to familiarize myself with the codebase, even if someone else has the feature ready. Glad to submit a PR if it's still needed, though!

- Yaniv (AKA Hansolite on Discord)

lily-pytel:
@YSavir no one has claimed it yet, so feel free to assign yourself to it and set it to in progress. Thank you for helping!

YSavir:
@lily-pytel Great! The assignees section doesn't allow me to add myself--is there another way I can go about that?

lily-pytel:
@YSavir added you!

alexmiddeleer:
@YSavir What's your discord handle if you want the contributor designation?

PR #219, opened 2019-11-11
Top level links to sub pages

YSavir:
Implements issue #204 

This PR allows setting `PageBanner` components as links, by wrapping the banner in an `a` tag. This can be done by passing a `url="<path>"` prop to the PageBanner component, which will then link to the given path.

This PR also adds the link to all sub-pages (eg. `/card`, `/event`) but not the top-level pages (eg. `/cards`, `/events`). This was done to be as true as possible to the ticket specifications, but assigning similar functionality to the top-level pages would be a quick effort. Let me know if I should add those.

This feature is implemented by checking for the `url` prop and invoking the correct render sub-function. This seemed the easiest way to implement, but if there's a better and more conventional method, let me know and I will update. I've worked with React before, but the React projects I've inherited didn't provide the best examples to approach such a situation. Creating two distinct PageBanner components seemed like overkill.

Originally I did try to extracting the shared styling to a helper function, but styled-jsx didn't know how to interpret that, unfortunately, so the styling is repeated.

(P.S. If there's a better branch to merge into other than Master, I'll re-open with that branch as the target)

### DIFF
--- a/next/components/page-banner.js
+++ b/next/components/page-banner.js
@@ -2,35 +2,67 @@ import { useContext } from 'react';
 import { ThemeContext } from './theme-context';
 import PropTypes from 'prop-types';
 
-function PageBanner({ children, image }) {
+function PageBanner({ children, image, url }) {
+  return (
+    (typeof(url) !== 'undefined' || url !== null )
+    ? _bannerWithUrl(children, image, url)
+    : _bannerWithoutUrl(children, image)
+  );
+}
+
+function _bannerWithoutUrl(children, image){
   const theme = useContext(ThemeContext);
+
   return (
     <div className="page-banner">
-      <style jsx>{`
-        .page-banner {
-          border-top: ${theme.border};
-          border-bottom: ${theme.border};
-          background: url(${image}) left no-repeat, ${theme.background};
-          height: 70px;
-          display: flex;
-          flex-direction: row-reverse;
-          align-items: center;
-        }
-        .page-banner h1 {
-          margin: 0;
-          padding-right: 0.5em;
-          text-align: right; /* multiline situations */
-        }
-
-        @media only screen and (max-width: 600px) {
-          .page-banner {
-            background: unset;
-          }
-        }
-      `}</style>
+      { _bannerStyling(image) }
       <h1 data-cy="header">{children}</h1>
     </div>
   );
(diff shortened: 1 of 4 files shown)


alexmiddeleer:
Thanks so much for the PR. Master is a fine. We'll test and code review this as soon as we have time, should be the next 48 hours usually. A few cypress tests wouldn't hurt, but is not required. If you haven't used cypress, it's extremely easy.

YSavir:
@alexmiddeleer I haven't used Cypress, but I'm always up for learning a new tool. Would love to see how it compares to Selenium scripts. I'll get that in tomorrow.

next/components/page-banner.js
```diff
@@ -1,16 +1,26 @@
...
 import { ThemeContext } from './theme-context';
 import PropTypes from 'prop-types';
+import Link from 'next/link';
 
-function PageBanner({ children, image }) {
+function PageBanner({ children, image, url }) {
+  return (
+    (typeof(url) !== 'undefined')
+    ? _bannerWithLink(children, image, url)
+    : _bannerWithoutLink(children, image)
```
alexmiddeleer: Should two functions just be extracted into components, do you think, rather than being functions in this file?

alexmiddeleer:
Thanks for doing this, going to accept all changes and give you the discord role of contributor. Thanks!

next/components/page-banner.js
YSavir: It's been merged in already, but I just saw this, and thought I'd at least give my thoughts.

I think it ultimately depends on what role Components play in this codebase. Personally, I like my components to each have a "role", and to allow for variance within that role (eg. link vs non-link). We could certainly break it out into two distinct components (or a parent and two sub-components), but I feel like that adds file bloat and distribute pieces of code that are relevant to each other into separate files. In my experience, while possibly consistent with some trends or other components in the codebase, it isn't necessarily improving the maintainability of the code--though future needs may prove it the opposite, of course.

But, if other components have received that treatment and consistency is paramount, it would be sensible to maintain that treatment in this case as well.

alexmiddeleer:
Well, I certainly see the logic in that. I guess they’re a bit like private
functions unless they’re needed in other places. Thanks for explaining!

PR #223, opened 2019-11-14
Show deck votes in the decks page

lily-pytel:
Like so

image


PapaCapricorn:
@lily-pytel This looks fantastic to me, and thank you for making the format consistent with the upvote styling on the other pages. Ultimately we might decide to move the rating column to a different position but that can wait until further discussion.

PR #230, opened 2019-11-17
Deck builder publish mode - part 2 - SQL + dropdowns + deck list

lily-pytel:
Some more work for deck publishing

- 2 enum types added
- 2 columns added
- deck preview modified
- migration file to reflect the above
- archetype and type added to deck queries
- archetype added to deck list
- dropdowns added and changes get saved to deck in progress

TODO:
- add type and archetype to deck page
- actually save them to the database
- and more


lily-pytel:
@alexmiddeleer adding those columns really messed up the sorting, though. Getting an error `structure of query does not match function result type` from postgres. Investigating, but boy... SQL functions are hard to decipher.

lily-pytel:
@alexmiddeleer I guess this was the problem: https://github.com/mythgard-hub/mythgard-hub/pull/230/commits/fbd888f26bb4f236a831fcb12321c8ffa1b939e6

PapaCapricorn:
@lily-pytel FYI I have added formatting for Types to the the deck page mockup (where previously only Archetypes were displayed). 

Font-weights are 500 for Archetype and 300 for Type, with each on separate lines. Please let me know if you need anything else from me. 

image

next/components/deck-list.js
```diff
@@ -41,6 +45,10 @@ export default function DeckList({ decks }) {
             const deckModifiedMeta =
               deckMetaData[index] && deckMetaData[index].deckCreated;
             const modified = new Date(deck.modified || deckModifiedMeta);
+            const archetype =
+              deckMetaData[index] && deckMetaData[index].deckArchetype
```
alexmiddeleer: deckMetaData[index] could probably be extracted to a variable at this point.. `metaData` maybe

alexmiddeleer: My brain's been a bit poisoned by Java lately and it's tempting to actually make a Deck class (like OO style class) where we put safe getters and setters. It would automatically handle stuff like checking if it has metadata, and default values. With ES6 style classes this doesn't seem like the PITA it once was.

This doesn't need to be in this PR... but do you have any thoughts about it?

lily-pytel: Extracted to variable.

As for classes, in my head I hear a howling of pain at the thought, but I don't have a good argument for this at the moment. I am not enjoying Java\Kotlin very much and my one experience with JS classes didn't feel like any improvement. BUT, this is not a good enough argument. Need to do some research.

PR #233, opened 2019-11-22
Moderators page (part 1 of descriptions)

alexmiddeleer:
Next PR will allow regular users to update their own decks' descriptions


express/whitelisted-queries.js
```diff
@@ -522,5 +522,26 @@ query tournaments($now: Date) {
       }
     }
   }
+`,
+  `
+  mutation updateDeck($deckId: Int!, $deckDesc: String, $deckName: String) {
+    updateDeck(
```
lily-pytel: Not `updateDeckDescription`?

next/pages/mod.js
```diff
@@ -0,0 +1,68 @@
...
+    const isModerator = data.accountModerators.totalCount;
+    if (isModerator) {
+      result = (
+        <div>
+          <h1>Mythgard moderators</h1>
+          <ModeratorControlPanel modUser={user} />
+        </div>
+      );
+    } else {
+      useEffect(() => {
```
lily-pytel: Does this not give you the error about rendering a different number of hooks between renders? Or does this error actually never happens. Also does it need an effect or is it just `error403()`?

lily-pytel:
Is the idea to enter my user id in the moderator table?

next/pages/mod.js
alexmiddeleer: It worked for me I think, but I'll double check. It does need an effect because we're redirecting and that fails on SSR. It's the recommended approach from the next peeps afaik.

express/whitelisted-queries.js
alexmiddeleer: It just a general patch. It also updates the name. In the future it might update more.

PR #236, opened 2019-11-23
Archetype type filter

lily-pytel:
@alexmiddeleer everything seems to be laid out correctly to do the filtering, but the function doesn't actually filter. I've hit a bit of a roadblock. It looks like cards get filtered through graphql as opposed to a function. I've also tried running it in adminer and it doesn't filter properly, so I think the issue is with the changes I added.

This doesn't filter
```
select mythgard.search_decks_nosort(NULL,NULL,NULL,NULL,NULL,NULL,NULL,NULL,NULL,NULL,NULL,NULL,NULL,NULL,NULL,array['MIDRANGE']::mythgard.deckArchetype[], NULL);
```

But this works:
```
select * from mythgard.deck where archetype = array['MIDRANGE']::mythgard.deckArchetype[]
```

It's tempting to make this a text field or filter on the front end, I won't lie. Any ideas?

### DIFF
--- a/next/components/deck-search-form-dropdown-filter.js
+++ b/next/components/deck-search-form-dropdown-filter.js
@@ -6,7 +6,6 @@ export default function DeckSearchFormDropdownFilter(props) {
 
   return (
     <label className="input-label">
-      <style jsx>{``}</style>
       {label}
       <br />
       <select
--- a/next/components/deck-search-form-dropdown-filter.js
+++ b/next/components/deck-search-form-dropdown-filter.js
@@ -1,0 +1,32 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export default function DeckSearchFormDropdownFilter(props) {
+  const { label, options, onChange, filterValue } = props;
+
+  return (
+    <label className="input-label">
+      <style jsx>{``}</style>
+      {label}
+      <br />
+      <select
+        className="filter-dropdown"
+        value={filterValue}
+        onChange={onChange}
+      >
+        {options.map(o => (
+          <option key={o.label} value={o.label}>
+            {o.label}
+          </option>
+        ))}
+      </select>
+    </label>
+  );
+}
+
+DeckSearchFormDropdownFilter.propTypes = {
+  label: PropTypes.string,
+  options: PropTypes.array,
+  filterValue: PropTypes.string,
+  onChange: PropTypes.func
+};
--- a/next/components/deck-search-form-dropdown-filter.js
+++ b/next/components/deck-search-form-dropdown-filter.js
@@ -14,6 +14,9 @@ export default function DeckSearchFormDropdownFilter(props) {
         value={filterValue}
         onChange={onChange}
       >
+        <option key="any" value={null}>
+          Any
+        </option>
         {options.map(o => (
           <option key={o.label} value={o.label}>
             {o.label}
--- a/next/components/deck-search-form-dropdown-filter.js
+++ b/next/components/deck-search-form-dropdown-filter.js
@@ -14,7 +14,7 @@ export default function DeckSearchFormDropdownFilter(props) {
         value={filterValue}
         onChange={onChange}
       >
-        <option key="any" value={null}>
+        <option key="any" value="">
           Any
         </option>
         {options.map(o => (
--- a/next/components/deck-search-form-dropdown-filter.js
+++ b/next/components/deck-search-form-dropdown-filter.js
@@ -11,6 +11,7 @@ export default function DeckSearchFormDropdownFilter(props) {
       <br />
       <select
         className="filter-dropdown"
+        data-cy={`deckSearch${label}`}
         value={filterValue}
         onChange={onChange}
       >
--- a/e2e/cypress/integration/deck-page.js
+++ b/e2e/cypress/integration/deck-page.js
@@ -7,8 +7,8 @@ describe('Deck Page', function() {
     cy.get('[data-cy="userLink"]').should('not.be.visible');
     cy.get('[data-cy="deckPageType"]').should('be.visible');
     cy.get('[data-cy="deckPageArchetype"]').should('be.visible');
-    cy.get('[data-cy="deckPageType"]').should('contain', 'standard');
-    cy.get('[data-cy="deckPageArchetype"]').should('contain', 'unknown');
+    cy.get('[data-cy="deckPageType"]').should('contain', 'Standard');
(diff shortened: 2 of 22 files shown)


lily-pytel:
@alexmiddeleer I made this test function
```
create or replace function mythgard.test_filter(
  archetype mythgard.deckArchetype[])
  returns setof mythgard.deck as $$

    SELECT deck.* FROM mythgard.deck
    WHERE deck.archetype = archetype::mythgard.deckArchetype[]

    GROUP BY deck.id
    LIMIT 2000;
  $$ language sql stable;

select mythgard.test_filter(array['MIDRANGE']::mythgard.deckArchetype[]);
```

to try to isolate it and it just doesn't work... it returns all rows.

alexmiddeleer:
I can take a look

alexmiddeleer:
@lily-pytel I'd suggest making the input a string and trying the `any` operator, perhaps like this.
```
create or replace function mythgard.test_filter(archy Text)
  returns setof mythgard.deck as $$

    SELECT deck.* FROM mythgard.deck
    WHERE archy::mythgard.deckArchetype = any(deck.archetype)

    GROUP BY deck.id
    LIMIT 2000;
  $$ language sql stable;

select mythgard.test_filter('MIDRANGE');
```
See https://www.postgresql.org/docs/current/arrays.html -> Searching in Arrays

lily-pytel:
@alexmiddeleer The problem is that it gets MIDRANGE, but also {CONTROL, MIDRANGE} - still, it's already better than before, when it wasn't working at all. Also how would you search for {CONTROL, MIDRANGE} in this scenario?

lily-pytel:
@alexmiddeleer Weirdly enough, this works:

```
DROP FUNCTION IF EXISTS filter_enum;
DROP TABLE IF EXISTS someTable;
DROP TYPE IF EXISTS enumExample;

CREATE TYPE enumExample AS ENUM ('blue', 'red', 'yellow', 'green');

CREATE TABLE someTable (
  id SERIAL PRIMARY KEY,
  enumColumn enumExample[]
);

INSERT INTO someTable (id, enumColumn) VALUES (1, array['blue']::enumExample[]);
INSERT INTO someTable (id, enumColumn) VALUES (2, array['blue', 'red']::enumExample[]);
INSERT INTO someTable (id, enumColumn) VALUES (3, array['red', 'blue']::enumExample[]);
INSERT INTO someTable (id, enumColumn) VALUES (4, array['blue', 'green']::enumExample[]);
INSERT INTO someTable (id, enumColumn) VALUES (5, array['red', 'yellow']::enumExample[]);

CREATE OR REPLACE function filter_enum(
  enumArray enumExample[])
  returns setof someTable as $$

    SELECT * FROM someTable
    WHERE someTable.enumColumn = enumArray::enumExample[]

  $$ language sql stable;

select filter_enum(array['blue']::enumExample[]);
select filter_enum(array['blue', 'red']::enumExample[]);
select filter_enum(array['green']::enumExample[]);
```

But this still doesn't

```
create or replace function mythgard.test_filter(
  archetype mythgard.deckArchetype[])
  returns setof mythgard.deck as $$

    SELECT * FROM mythgard.deck
    WHERE deck.archetype = archetype::mythgard.deckArchetype[]

  $$ language sql stable;

select mythgard.test_filter(array['MIDRANGE']::mythgard.deckArchetype[]);
select mythgard.test_filter(array['CONTROL', 'MIDRANGE']::mythgard.deckArchetype[]);
```

WHAT IS THE DIFFERENCE???

alexmiddeleer:
I figured we don’t ever need to allow them to search for multiple.

lily-pytel:
@alexmiddeleer Ugh, I found the difference... and it's so stupid I want to cry: https://github.com/mythgard-hub/mythgard-hub/pull/236/commits/74cb562413f3700ddd9f82c140964a652ade3d2b

Gonna add some test for the new filtering and then this feature should be good to go. Only last annoyance is that the deck page doesn't reload properly after an update but that was happening before as well

alexmiddeleer:
Sql is really tricky sometimes.

lily-pytel:
@alexmiddeleer okay, it's ready for review. I'll test everything again once it's merged to the main feature branch. Thanks for looking into it!

next/components/deck-search-form-dropdown-filter.js
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export default function DeckSearchFormDropdownFilter(props) {
+  const { label, options, onChange, filterValue } = props;
+
+  return (
+    <label className="input-label">
+      <style jsx>{``}</style>
```
alexmiddeleer: ✂️ ?

next/components/deck-search-form-dropdown-filter.js
```diff
@@ -0,0 +1,36 @@
...
+      {label}
+      <br />
```
alexmiddeleer: Is this br needed or would `display: block` be better? I view br tags as a slightly less odious version of inline styles, but they're convenient enough that I wouldn't blacklist them everywhere.

alexmiddeleer:
`Warning: `value` prop on `select` should not be null. Consider using an empty string to clear the component or `undefined` for uncontrolled components. `

Gettings this after navigation from /home to /decks on first attempt

next/components/deck-search-form-dropdown-filter.js
lily-pytel: It's what we do for other parts of this page, I was mostly trying to be consistent there

next/components/deck-search-form-dropdown-filter.js
lily-pytel: I was wondering if you would say that. Will remove.

lily-pytel:
> Warning: value prop on select should not be null. Consider using an empty string to clear the component or undefined for uncontrolled components.

I hate this warning a lot. Null is the best value to use there and I wish they would just leave it alone. Will change though.

next/components/deck-search-form-dropdown-filter.js
alexmiddeleer: Cool, we can leave it in

alexmiddeleer:
This error is the only remaining issue

image

lily-pytel:
@alexmiddeleer I think you're missing the last commit. It looks like the old error message - that line isn't the same after https://github.com/mythgard-hub/mythgard-hub/pull/236/commits/0c0939cce9293c069ea2f7c1102215423bec31a7

PR #238, opened 2019-11-25
Deck descriptions (final)

alexmiddeleer:


lily-pytel:
Though I guess you can now change the base to master and pull from there. Should be the same changes, but the `deck-builder-publish-mode` branch is obslete now

PR #241, opened 2019-11-26
Add arena type

lily-pytel:
- Added arena type in ui
- Added arena type in db
- Renamed previous migrations file
- Added new migration file

NOTE: @alexmiddeleer while the migration worked for me on the database, postgraphile did not accept the new type. It might need to be restarted.


alexmiddeleer:
Lgtm! I can’t approve on a phone. Thanks!

PR #242, opened 2019-11-27
Card count deck page

lily-pytel:
I added the images to the Trello cards and forgot to save them. Excuse the laziness:

card count
https://trello.com/c/kSIrzn2g/225-deck-page-add-card-count

deck page mobile
https://trello.com/c/nBzOarrN/226-deck-page-make-one-column-on-mobile

### DIFF
--- a/next/components/deck.js
+++ b/next/components/deck.js
@@ -13,7 +13,8 @@ import {
   getDateCreated,
   getFactions,
   getDeckArchetype,
-  getDeckType
+  getDeckType,
+  getCardCount
 } from '../lib/deck-utils';
 import DeckVote from './deck-vote';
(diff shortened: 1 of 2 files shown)


next/components/deck.js
```diff
@@ -131,21 +133,37 @@ export default function Deck({ deck }) {
...
+          display: flex;
+          justify-content: space-between;
+        }
+
         @media only screen and (max-width: 575.98px) {
           .deck-page-container {
             flex-direction: column;
           }
+          .two-columns {
+            flex-direction: column;
```
alexmiddeleer: This totally works, although imposing a minimum-width on the second column would have also worked to force the stacking to occur and would not have required using the media query. It's doesn't make much difference though.

PR #243, opened 2019-11-29
Editable config

alexmiddeleer:
Allows mods to update media (e.g. articles). More importantly, is a POC for the two new hooks that control getting and setting the site config, a json document that now lives in postgres that we can use for anything from the banner message, to advertisements, etc.
image


next/components/moderator-config-editor.js
```diff
@@ -0,0 +1,75 @@
+import useConfig from '../lib/use-config.js';
+import useConfigMutation from '../lib/use-config-mutation.js';
+import ModeratorEditArticle from '../components/moderator-edit-article.js';
+
+function arrayMove(arr, fromIndex, toIndex) {
```
lily-pytel: 🤦‍♂ that we still have to do stuff like this in 2019

next/components/moderator-config-editor.js
```diff
@@ -0,0 +1,75 @@
+import useConfig from '../lib/use-config.js';
```
lily-pytel: So is this a way to hide the useQuery and useMutation stuff? What's happening here?

alexmiddeleer:
Not only that but hiding the serialization and deserialization of the json
and extracting the only field we care about from the response object.

PR #245, opened 2019-11-30
Deck mana curve using our old friend highcharts

lily-pytel:
- Added highcharts to the project. I have tried others but honestly highcharts still rules.
- Added a deck-stats utility file to create the mana curve series + unit tests
- Added mana curve chart to the deck page under factions
- Added mana curve charts to the deck builder, under the action buttons - if there are any cards
- Tiny e2e tests.

### DIFF
--- a/next/components/deck-mana-curve.js
+++ b/next/components/deck-mana-curve.js
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import Highcharts from 'highcharts';
 import HighchartsExporting from 'highcharts/modules/exporting';
 import HighchartsReact from 'highcharts-react-official';
-import { getManaCurve } from '../lib/deck-stats';
+import { getManaCurveHighchartsSeries } from '../lib/deck-stats';
 import { ThemeContext } from './theme-context';
 
 if (typeof Highcharts === 'object') {
@@ -60,7 +60,7 @@ export default function DeckManaCurve({ cards }) {
         stacking: 'normal'
       }
     },
-    series: getManaCurve(cards, theme)
+    series: getManaCurveHighchartsSeries(cards, theme)
   };
 
   return (
--- a/next/components/deck-mana-curve.js
+++ b/next/components/deck-mana-curve.js
@@ -1,15 +1,17 @@
+import { useContext } from 'react';
 import PropTypes from 'prop-types';
 import Highcharts from 'highcharts';
 import HighchartsExporting from 'highcharts/modules/exporting';
 import HighchartsReact from 'highcharts-react-official';
 import { getManaCurve } from '../lib/deck-stats';
+import { ThemeContext } from './theme-context';
 
 if (typeof Highcharts === 'object') {
   HighchartsExporting(Highcharts);
 }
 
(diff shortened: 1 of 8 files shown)


next/components/deck-mana-curve.js
```diff
@@ -0,0 +1,78 @@
...
+if (typeof Highcharts === 'object') {
+  HighchartsExporting(Highcharts);
+}
+
+export default function DeckManaCurve({ cards }) {
+  const theme = useContext(ThemeContext);
+  const options = {
+    chart: {
+      type: 'column',
+      backgroundColor: 'rgba(0,0,0,0)',
```
lily-pytel: I'm honestly not sure how much to modularize this business. I feel like it's pretty locked in since it only exists in this component but I can also see making more helper functions and stuff to create this config... tell me your thoughts.

alexmiddeleer: It's fine with me to keep it simple for now

next/components/deck-mana-curve.js
```diff
@@ -0,0 +1,78 @@
+import { useContext } from 'react';
+import PropTypes from 'prop-types';
+import Highcharts from 'highcharts';
+import HighchartsExporting from 'highcharts/modules/exporting';
```
alexmiddeleer: Do we need to support this?

next/lib/deck-stats.js
```diff
@@ -0,0 +1,70 @@
...
+    case FACTION_COLORS.orange:
+      color = theme.orangeFactionColor;
+      break;
+    case FACTION_COLORS.green:
+      color = theme.greenFactionColor;
+      break;
+  }
+
+  return color;
+};
```
alexmiddeleer: You can do this without a switch if you define a constant object, e.g.

```
const FACTION_COLORS_MAP = {};

const factionMainColor = (faction, theme) => { 
    // This step can be memoized so it runs once..
    FACTION_COLORS_MAP[FACTION_COLORS.green]  = theme.greenFactionColor;
    ... etc

   // Then just return this step every time.
   return FACTION_COLORS_MAP[faction];
}
```

next/lib/deck-stats.js
```diff
@@ -0,0 +1,70 @@
...
+
+const initializeManaCostsByFunction = theme => {
```
alexmiddeleer: `initializeManaCostsByFaction`?

next/components/deck-mana-curve.js
lily-pytel: It's what their docs say to use for server side rendering. I admit I didn't get super into this beyond this part in their docs: https://github.com/highcharts/highcharts-react#highcharts-with-nextjs

next/lib/deck-stats.js
lily-pytel: Oopd

next/lib/deck-stats.js
lily-pytel: True

lily-pytel: I copied this from some other place 😟

PR #248, opened 2019-12-04
Mana curve x cost fix

### LINKED ISSUE #246: Wrapping columns
I believe these shouldn't be allowed to wrap (decks page table). A simple white-space: nowrap should fix this I imagine.

image

lily-pytel:
- Associate -1 cost cards as 1 cost cards (but without abs)
- Remove the ability for most cells in the decks table to wrap - fixes #246

### DIFF
--- a/next/lib/deck-stats.js
+++ b/next/lib/deck-stats.js
@@ -42,7 +42,16 @@ export const getManaCurveHighchartsSeries = (cards, theme) => {
     Object.values(cards).forEach(c => {
       const { quantity, card } = c;
       const primaryFaction = mainFaction(card);
-      const manaCost = card.mana >= 6 ? '6+' : card.mana.toString();
+
+      let manaCost = card.mana;
+      if (manaCost === -1) {
+        manaCost = '1';
+      } else if (manaCost >= 6) {
+        manaCost = '6+';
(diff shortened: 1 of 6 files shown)


next/lib/deck-stats.js
```diff
@@ -42,7 +42,16 @@ export const getManaCurveHighchartsSeries = (cards, theme) => {
     Object.values(cards).forEach(c => {
       const { quantity, card } = c;
       const primaryFaction = mainFaction(card);
-      const manaCost = card.mana >= 6 ? '6+' : card.mana.toString();
+
+      let manaCost = card.mana;
+      if (manaCost === -1) {
```
lily-pytel: Too explicit?

alexmiddeleer: No. I might even make it more explicit by adding a constant X_COST

lily-pytel: Can't tell if you're kidding

lily-pytel: But here you go @alexmiddeleer : https://github.com/mythgard-hub/mythgard-hub/pull/248/commits/654674fe0a6223fa2ec7cc937741156169d608e3

PR #255, opened 2019-12-21
Fix faction order

lily-pytel:
Not super sure why we needed the reduce and unshift... seems like this simple map should do it?

### DIFF
--- a/next/components/card.js
+++ b/next/components/card.js
@@ -36,7 +36,7 @@ export default function Card({ card }) {
   }
 
   const primaryFaction = factions.length ? factions[0] : '';
-  const secondaryFaction = factions.length === 2 ? factions[1] : '';
+  const secondaryFaction = factions.length > 1 ? factions[1] : '';
 
   return (
     <>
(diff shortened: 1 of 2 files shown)


next/components/card.js
```diff
@@ -18,25 +18,26 @@ export default function Card({ card }) {
...
+
   try {
-    factions = card.cardFactions.nodes.reduce(
-      (acc, n) => {
-        if (n && n.faction && n.faction.name) {
-          acc.unshift(firstLetterUppercase(n.faction.name));
-        }
-        return acc;
-      },
-      ['', '']
```
alexmiddeleer: Looks like we wanted to to make sure there were minimum two elements

next/components/card.js
```diff
@@ -18,25 +18,26 @@ export default function Card({ card }) {
...
   let spawns = [];
+
   try {
     spawns = card.cardSpawns.nodes.map(n => n.spawn.name);
   } catch (error) {
     console.error('Something went wrong trying to read card spawns', error);
   }
+
+  const primaryFaction = factions.length ? factions[0] : '';
+  const secondaryFaction = factions.length > 1 ? factions[1] : '';
```
alexmiddeleer: But, looks like that's now being handled here in a much simpler way, so 👍

PR #256, opened 2019-12-24
Deck builder plus minus

### LINKED ISSUE #240: Card types only appear after saving a deck
After saving a deck, the table shows the type near the amount:

image

However, when you just add cards to the deck - by importing or clicking on them - they don't have the type. It would be nice to have it there.

image

Steps to reproduce:
1. Save a deck
2. Edit the deck
3. Remove one of the cards
4. Add it back

lily-pytel:
- shows the card type in deck builder (fixes #240)
- adds a plus and minus button on hover for desktop
- adds a plus and minus button permanently on mobile
- adds unit tests. Sadly there is no hover capability in cypress so there's no great way to test this.

Desktop:
image

Mobile:
image

### DIFF
--- a/express/whitelisted-queries.js
+++ b/express/whitelisted-queries.js
@@ -177,6 +177,7 @@ mutation UpdateDeckAndRemoveCards(
         mana
         gem
         rarity
+        supertype
         cardFactions {
           nodes {
             faction {
@@ -241,6 +242,7 @@ mutation UpdateDeckAndRemoveCards(
           id
           mana
           gem
+          supertype
           cardFactions {
             nodes {
               faction {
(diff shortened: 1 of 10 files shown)


alexmiddeleer:
I might be crazy but is that plus italicized?

lily-pytel:
> I might be crazy but is that plus italicized?

@alexmiddeleer you're not crazy, that's our base button styling (italicized text). Should I unitalicize? I didn't touch it since I didn't want to mess with our base design - though I'm not a huge fan of the base italics in the buttons in general

alexmiddeleer:
We could just add a text style reset class or something. Might want to ask
the ux peeps.

PR #258, opened 2020-01-18
Update hotness

alexmiddeleer:
Basically this makes the number of votes less important and time more important.


alexmiddeleer:
I suppose it's relevant to say that this is already live.

PR #259, opened 2020-02-07
Fix plus sign craziness + add 0 cost cards to mana curve

lily-pytel:
It was just doubling itself... I don't know what I was thinking 🤦‍♂. If I remember correctly, I wasn't able to get the hover working at the time to test this, so probably why I didn't catch it in the test. Mea culpa!


lily-pytel:
@alexmiddeleer also are the tests going crazy for you too?

lily-pytel:
Definitely not my best naming work

PR #261, opened 2020-02-13
Filter for existing factions only

lily-pytel:
Javascript be javascripting. This doesn't happen on our local database because it's not as chaotic, I'm guessing?

### DIFF
--- a/next/components/card.js
+++ b/next/components/card.js
@@ -20,9 +20,9 @@ export default function Card({ card }) {
   let factions = [];
 
   try {
-    factions = card.cardFactions.nodes.map(f =>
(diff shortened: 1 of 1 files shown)


alexmiddeleer:
What's the background on this?

lily-pytel:
In the production database, the array has two factions but the second one is null, so the map doesn't work and defaults to []. In the local database, it doesn't happen

lily-pytel:
@alexmiddeleer

alexmiddeleer:
Should we fix it on to the db side?

lily-pytel:
@alexmiddeleer probably, but you know I am always in favor of paranoid javascript as well

PR #270, opened 2020-04-25
one way to fix the breaking changes

lilypytel-toast:
Fixes this breaking change that was implemented Mid Feb:
image

I'm still trying to figure out if this is okay. We can of course set a password as well

### DIFF
--- a/e2e/package.json
+++ b/e2e/package.json
@@ -4,7 +4,7 @@
   "description": "",
   "main": "index.js",
   "dependencies": {
-    "cypress": "^3.4.1"
+    "cypress": "^4.4.1"
   },
   "scripts": {
     "cypress:open": "cypress open"
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-  - "lts/*"
+  - '12.12.0'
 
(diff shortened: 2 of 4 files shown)


jtrussell:
What was going on?

lily-pytel:
@jtrussell postgres updated with a breaking change in February AND node also has something that doesn't work the same between versions. Local dev and CI were broken as a result 🤦 

Node: https://github.com/manuelbieh/geolib/issues/208 (this is just a discussion about the issue. Not related to travis or cypress but was where I got the idea to try and lock the node version. I actually can't find info related to cypress or travis)
Postgres: https://github.com/docker-library/postgres/issues/681

I don't feel great about adding `POSTGRES_HOST_AUTH_METHOD: 'trust'`, but it looks like that's what other people do to get local and CI working. It's not like this docker-compose is applied in production... right?

e2e/package.json
```diff
@@ -4,7 +4,7 @@
   "description": "",
   "main": "index.js",
   "dependencies": {
-    "cypress": "^3.4.1"
+    "cypress": "^4.4.1"
```
lily-pytel, line 7: This is not necessary but whatever, it didn't break

PR #271, opened 2020-04-26
Add path and power icon to the deck cards table

lily-pytel:
Adds the path and power icon to the table, if it exists: 
image

The code for getting the image is not very sophisticated - the names of the images don't really match the names of the cards so other than attaching them in the db, this seems like the only other reliable way. I thought about attaching them to the database, but is it worth it?


next/constants/powers.js
```diff
@@ -0,0 +1,29 @@
+const cdn = `${process.env.MG_CDN}/filters/`;
+
+export const matchPowerNameToIcon = power => {
+  try {
+    const powerName = power.name;
+    const lowerCaseName = powerName.toLowerCase();
+
+    switch (lowerCaseName) {
```
alexmiddeleer, line 8: could use a static object instead of a switch
```
    const powerIcons = {
       forsight: `${cdn}foresight_slug.png`,
     ...
    }
 
...
   export const matchPowerNameToIcon = ...
...
    return powerIcons[lowerCaseName] || null;
```

next/constants/paths.js
```diff
@@ -0,0 +1,25 @@
...
+      case 'disk of circadia':
+        return `${cdn}DiskOfCircadia.png`;
+      case 'fires of creation':
+        return `${cdn}FiresOfCreation.png`;
+      case 'journey of souls':
+        return `${cdn}JourneyOfSouls.png`;
+      case "rainbow's end":
+        return `${cdn}RainbowsEnd.png`;
+      case 'turn of seasons':
+        return `${cdn}TurnOfSeasons.png`;
```
alexmiddeleer, line 18: See suggestion for powerIcons

next/constants/powers.js
lily-pytel: Done

PR #272, opened 2020-04-29
Deck page views

lily-pytel:


express/whitelisted-queries.js
```diff
@@ -561,6 +562,13 @@ query tournaments($now: Date) {
       }
     }
   }
+`,
+  `
+  mutation increaseDeckViews($deckid: Int!) {
+    increaseDeckViews(input: { deckid: $deckid }) {
+      integer
```
lily-pytel, line 569: Cannot figure out how to call this something more specific. The general theme of this feature was "Is postgraphile stupid or am I stupid or are we both stupid?"

next/pages/deck.js
```diff
@@ -10,16 +11,49 @@ import {
...
+              ...deck,
+              views: newViews
+            }
+          }
+        });
+      }
+    }
+  );
+
+  useEffect(() => {
```
lily-pytel, line 48: I believe this is the best way to record a view but LMK if there's another idea

next/pages/deck.js
```diff
@@ -10,16 +11,49 @@ import {
...
+    increaseViews({
+      variables: {
+        deckid: deckId
```
lily-pytel, line 51: `deckid` versus `deckId` was another instance of "who is being stupid here?" I tried various combinations of `deckId` and `deck_id` but finally gave up and just made everything lower case in postgres and graphql :tableflip:

next/pages/deck.js
```diff
@@ -10,16 +11,49 @@ import {
...
+      }
+    });
+  }, []);
 
-  if (loading)
+  if (loading || viewsUpdating)
```
lily-pytel, line 56: I figure we don't want to see the page load and then the view increase, right?

postgres/init.sql
```diff
@@ -486,6 +487,17 @@ RETURNS INTEGER AS $$
...
+CREATE OR REPLACE FUNCTION mythgard.increase_deck_views(IN deckid INTEGER)
+RETURNS INTEGER AS $$
+
+  UPDATE mythgard.deck SET views = views + 1
+  WHERE deck.id = deckid;
+
+  SELECT views
+  FROM mythgard.deck
+  WHERE deck.id = deckid;
+$$ language sql VOLATILE SECURITY DEFINER;
```
lily-pytel, line 499: `VOLATILE` is apparently required for mutations but without `SECURITY DEFINER` it would do the SELECT but not the UPDATE. Should I instead grant a more specific permission?

postgres/migrations/migration-1.8.0.sql
```diff
@@ -0,0 +1,18 @@
+ALTER TABLE mythgard.deck_featured ENABLE ROW LEVEL SECURITY;
```
lily-pytel, line 1: Is this in fact the correct version for the last release? I just guessed

next/pages/deck.js
alexmiddeleer: I'd actually prefer the page load faster, I guess, but it doesn't matter much.

postgres/init.sql
alexmiddeleer: I'm not sure, to be honest.

postgres/migrations/migration-1.8.0.sql
alexmiddeleer: You can verify by looking at the git tags. They should line up. I should have created that file last time I deployed but it seems I forgot.

next/pages/deck.js
alexmiddeleer: To be honest I hadn't thought much about it. I think this makes the most sense, at least it's intuitive and won't have weird bugs where we made an extra XHR call and that accidentally bumped the views.

next/pages/deck.js
```diff
@@ -10,16 +11,56 @@ import {
...
+    increaseDeckViews,
+    {
+      update: (cache, { data }) => {
+        const newViews = data.increaseDeckViews.integer;
+        const { deck } = cache.readQuery({
+          query: singleDeckQuery,
+          variables: { id: deckId }
+        });
+
+        cache.writeQuery({
```
alexmiddeleer, line 34: this is one of my least favorite things about apollo client. `writeQuery` code is always terrifying. Do we not need to include the `_typename`s like I've had to do in other writeQuery's? It always errors in my console if I don't add that.

express/whitelisted-queries.js
alexmiddeleer: I'm always smashing my head against the wall when I mess with this stuff. I just have to play with it in graphiql and have the postgraphile docs open. Unfortunately I don't know the answer off the top of my head, you might need one of their "smart" comments. https://www.graphile.org/postgraphile/smart-comments/

next/pages/deck.js
alexmiddeleer: It's definitely a pain to get used to. And functions are harder than just CRUD on a table. All I can suggest is messing around in graphiql until it works the way you want then copy it over. I'm not sure off the top of my head what the issue is but I believe the variable names here are only related to the name of the param in the mutation, and have nothing to do with the actual postgres schema.

next/pages/deck.js
lily-pytel: I didn't notice any errors - it seems to work? But I'll look. At work I just do refetch queries instead of manually changing the cache because jfc

alexmiddeleer: That's fine too. I mean for this domain I probably would be fine with it being off-by-one (or even off-by-a-lot), meaning you could just skip the entire writeCache step if you want.

next/pages/deck.js
lily-pytel: It doesn't seem to make a notable difference loading wise, but I can just get rid of it. By the time someone scrolls to the bottom of the deck page, they probably won't have noticed this

postgres/init.sql
lily-pytel: That's what the docs said so I guessss I'll just leave it at that and we'll see what happens?

postgres/migrations/migration-1.8.0.sql
lily-pytel: Looks like the latest release is 1.7.6 in the but there is a file for 1.7.7 so I renamed this one 1.7.8

next/pages/deck.js
lily-pytel: It's cool, it's there now and it's consistent with other parts of the app

express/whitelisted-queries.js
lily-pytel: This is what I got after playing with it in graphiql and the docs 🙃 . That's how I realized it was returning `integer`. I didn't see how to apply smart comments in this case, or did I miss anything?

next/pages/deck.js
lily-pytel: After some distance to cool off, figured out the relationships between the variables and the function and the mutation and everything

express/whitelisted-queries.js
alexmiddeleer: I think there's a way to tell it the names of function parameters but whatever, if it works it's fine.

Issue #260, opened 2020-02-12
Three column layout for homepage


PapaCapricorn:
Modify the [homepage layout](https://mythgardhub.com/) to follow the three column layout depicted on the [homepage mockup](https://mythgard-hub.github.io/index.html)

- *Column 1*: Recent Media - media posts filtered by most recent (these are not rated, so 'Recent' makes more sense than 'Top')
- *Column 2*: New decks - decks filtered by more recently added
- *Column 3*: Top decks - decks filtered by most currently popular, using the Hot algorithim

Top banner should remain as current (new player guide), Inked Gaming banner is placeholder only

PR #273, opened 2020-05-16
3 column homepage

lily-pytel:
Desktop:
image

Looks a little strange on a smaller screen, but not too bad:
image

### DIFF
--- a/next/components/author-link.js
+++ b/next/components/author-link.js
@@ -1,0 +1,25 @@
+import { useContext } from 'react';
+import PropTypes from 'prop-types';
+import { ThemeContext } from './theme-context';
+
+export default function AuthorLink({ author }) {
+  const theme = useContext(ThemeContext);
+  const link = author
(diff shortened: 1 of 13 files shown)


lily-pytel:
Addresses #260

next/components/author-link.js
```diff
@@ -0,0 +1,25 @@
+import { useContext } from 'react';
+import PropTypes from 'prop-types';
+import { ThemeContext } from './theme-context';
+
+export default function AuthorLink({ author }) {
+  const theme = useContext(ThemeContext);
+  const link = author
+    ? `/decks?updatedTime=100000&authorName=${author}&sortBy=hot`
```
alexmiddeleer, line 8: Does updatedTime need to be included? Eventually this might cause a bug report 😄

lily-pytel: Otherwise it shows only the last 150 days 🤷

lily-pytel:
LMK whether we want to do something else with the updated param. Perhaps by the time it's a problem, we'll have profile pages.

PR #275, opened 2020-05-22
Order cards factions correctly

alexmiddeleer:


alexmiddeleer:
> Love the tests! They inspire a lot of confidence

A little more at least!

PR #277, opened 2020-06-27
Add new patreon tiers to the database

alexmiddeleer:


lilypytel-toast:
Looks good but I wonder wtf happened to Travis. Also have not been able to run this locally yet - I keep getting `column "accountType" of relation "account" does not exist` when building the db which... is not true, it's right there in the sql script

lily-pytel:
Looks good but I wonder wtf happened to Travis. Also have not been able to run this locally yet - I keep getting column "accountType" of relation "account" does not exist when building the db which... is not true, it's right there in the sql script

lily-pytel:
Names in postgres, seriously. It works when I rename it to `account_type` like `google_id`

alexmiddeleer:
Sigh. lol.

lily-pytel:
I made a branch off yours with a different name and some changes to the local data: https://github.com/mythgard-hub/mythgard-hub/pull/278 it's failing one of the Travis checks because of flakiness but it'll pass eventually

PR #278, opened 2020-06-27
PR suggestions

lily-pytel:
will this fix travis?

### DIFF
--- a/e2e/cypress.json
+++ b/e2e/cypress.json
@@ -1,3 +1,4 @@
 {
-  "baseUrl": "http://localhost:3001"
+  "baseUrl": "http://localhost:3001",
+  "defaultCommandTimeout": 10000
 }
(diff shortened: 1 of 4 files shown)


lily-pytel:
At least on my machine, postgres or whoever has a problem with the camel case and refuses to acknowledge it, _even though_
 I read that putting the name in quotes should have been fine. Honestly this is typical of postgres, I feel.

lily-pytel:
Travis times out on the decks test, but unclear if it's real or flaky. Investigating

lily-pytel:
Hmm, it's having the same problem where `cats` is the first deck when you go to `http://localhost:3001/decks` and `dragons` is the first deck when you actually sort by hot and go to `http://localhost:3001/decks?name=&cardIds=&factionNames=&isOnlyFactions=true&updatedTime=150&archetype=&type=&authorName=&sortBy=hot`. 

Last time that was happening because cats was being updated by adding the views. In this case the decks are seemingly not changed at all because we're just adding data to unrelated tables.

lily-pytel:
Welp, dragons and cats have the same hotness value so when sorting by hotness, cats happen to be first, but when sorting by hotness _and_ updated time, dragons is first because it does get updates some miliseconds later. Jeez. I made dragons hotter by giving it more votes so it should work now. Alternatively we could break hotness ties by votes or dates modified but w\e, this probably only happens in our test data

PR #280, opened 2020-07-03
change author link to point to the account page

lily-pytel:


alexmiddeleer:
Nice, glad that component was there already

PR #281, opened 2020-07-03
Public acct page

alexmiddeleer:


postgres/migrations/next-migration.sql
```diff
@@ -1,11 +1,13 @@
...
+alter table mythgard.account add column profile_icon_id integer default 1;
 
 CREATE POLICY update_account_if_moderator
   ON mythgard.account
   FOR UPDATE
   USING (exists(select * from mythgard.account_moderator
          where account_id = mythgard.current_user_id()));
 
-GRANT UPDATE (account_type) ON TABLE mythgard.account TO authd_user;
+GRANT UPDATE (account_type, profile_icon_id) ON TABLE mythgard.account TO authd_user;
```
alexmiddeleer, line 12: not yet used.

alexmiddeleer:
I should write a cypress test for this I think, at least to check that the page loads.

next/components/profile.js
```diff
@@ -0,0 +1,57 @@
...
+      <style jsx>{`
+        .user-profile {
+          text-align: center;
+        }
+        .profile-image {
+          width: 138px;
+          margin-top: 30px;
+        }
+        .user-name {
+          color: #f1810b;
```
lily-pytel, line 25: Any reason not to use our theme?

alexmiddeleer: Uh, none - I just copy pasted this from the old page.

alexmiddeleer: Fixed, good call.

PR #284, opened 2020-07-09
Set profile icon

alexmiddeleer:


next/components/avatar-picker.js
```diff
@@ -0,0 +1,116 @@
...
+// import { useContext } from 'react';
+// import { ThemeContext } from './theme-context';
+import PropTypes from 'prop-types';
+
+const numAvatars = 62;
+const getLink = id => `${cdn}/avatars/avatar-${id}.png`;
+const cdn = process.env.MG_CDN;
+
+const limits = {
+  default: 31,
```
lily-pytel, line 11: I think this needs to be `basic`. It looks like the basic account has access to all avatars 😆

lily-pytel: Also I would just make these upper case so as not to even bother with the `toLowerCase` call

alexmiddeleer:
@lily-pytel Thanks for finding that bug 🙇

PR #287, opened 2020-07-19
Prettier public profile page

alexmiddeleer:


next/components/user-top-decks.js
```diff
@@ -0,0 +1,73 @@
+import PropTypes from 'prop-types';
```
alexmiddeleer, line 1: To be honest this is basically a copy paste of user-newest-decks. Maybe I should DRY it up... thoughts?

lily-pytel: Always!

lily-pytel: It looks kind of the same as the homepage lists too so you can even further

alexmiddeleer: Good points. I made some new components to DRY it up; lmk if they look alright.

next/components/view-more.js
```diff
@@ -0,0 +1,46 @@
+import PropTypes from 'prop-types';
+
+function ViewMore({ url, count, limit }) {
```
lily-pytel, line 3: You can just put the limit here, since I believe `defaultProps` are going away

alexmiddeleer:
> The code and profile page look great! I would just take out the underline in the view more link but it's whatever.
> image

Good catch on the underline. Might as well make it look like we're trying to match the design at least a bit, heh.

PR #290, opened 2020-08-02
Deck search by path power

lily-pytel:
The filters look like this (like our usual):
image

The paths and powers themselves can't quite be seen since we don't have images for our fake ones, but if you import the prod database and run the migrations, it looks like this
image

### DIFF
--- a/express/whitelisted-queries.js
+++ b/express/whitelisted-queries.js
@@ -7,6 +7,8 @@ const deckPreviewsFragment = `
     votes
     deckArchetype
     deckType
+    pathName
+    powerName
     views
     deck{
       id
@@ -310,6 +312,8 @@ mutation UpdateDeckAndRemoveCards(
         votes
         deckArchetype
         deckType
+        pathName
+        powerName
       }
     }
   }
--- a/express/whitelisted-queries.js
+++ b/express/whitelisted-queries.js
@@ -412,6 +412,8 @@ mutation UpdateDeckAndRemoveCards(
       $numFactions: Int
       $archetype: [Deckarchetype]
       $type: [Decktype]
+      $pathName: String
+      $powerName: String
       $first: Int
       $offset: Int
       $sortBy: String
@@ -434,6 +436,8 @@ mutation UpdateDeckAndRemoveCards(
         numfactions: $numFactions
         archetypefilter: $archetype
         typefilter: $type
+        pathname: $pathName
+        powername: $powerName
         first: $first
         offset: $offset
         sortby: $sortBy
@@ -691,6 +695,22 @@ mutation addFeaturedDeck($deckId: Int!) {
     ) {
       ${deckPreviewsFragment}
       totalCount
+    }
+  }
+`,
+  `
+  query pathPowerQuery {
+    paths {
+      nodes {
+        id
+        name
+      }
+    }
+    powers {
+      nodes {
+        id
+        name
+      }
     }
   }
 `,
--- a/next/components/deck-search-form.js
+++ b/next/components/deck-search-form.js
@@ -214,6 +214,7 @@ export default function DeckSearchForm(props) {
             )}
             onSubmit={handleSubmit}
           />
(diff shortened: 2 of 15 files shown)


express/whitelisted-queries.js
```diff
@@ -689,5 +697,21 @@ mutation addFeaturedDeck($deckId: Int!) {
       totalCount
     }
   }
+`,
+  `
+  query pathPowerQuery {
```
lily-pytel, line 702: TBH this query is mostly so we can use it with our test data

next/components/deck-search-form.js
```diff
@@ -214,6 +239,16 @@ export default function DeckSearchForm(props) {
             )}
             onSubmit={handleSubmit}
           />
+          {paths && paths.length && (
```
lily-pytel, line 242: if we fail to get this from the db for some reason just don't show this filter instead of bombing the whole page... thoughts?

postgres/migrations/next-migration.sql
```diff
@@ -0,0 +1,244 @@
...
+
+    SELECT deck.* FROM mythgard.deck
+      LEFT JOIN mythgard.card_deck ON (card_deck.deck_id = deck.id)
+      LEFT JOIN mythgard.card ON (card.id = card_deck.card_id)
+      LEFT JOIN mythgard.account ON (account.id = deck.author_id)
+      LEFT JOIN mythgard.path ON (path.id = deck.path_id)
+      LEFT JOIN mythgard.power ON (power.id = deck.power_id)
+
+    -- deck name filter
+    WHERE (deckName is NULL or trim(deckName) = '' or to_tsvector('simple', deck.name) @@ to_tsquery('simple', replace(regexp_replace(trim(deckName), '\s+', ' ', 'g'), ' ', ':* & ') || ':*'))
```
lily-pytel, line 79: Note that when I try to run this in adminer, I get an error but this is old code so I'm guessing it's an adminer issue... hopefully

postgres/migrations/next-migration.sql
```diff
@@ -0,0 +1,244 @@
...
+  ON mythgard.account.id = mythgard.deck.author_id
+  LEFT JOIN mythgard.deck_views
+  ON mythgard.deck_views.deck_id = mythgard.deck.id
+  LEFT JOIN mythgard.path
+  ON mythgard.path.id = mythgard.deck.path_id
+  LEFT JOIN mythgard.power
+  ON mythgard.power.id = mythgard.deck.power_id
+;
+
+drop function if exists mythgard.search_decks_nosort(
```
lily-pytel, line 30: I remember you have to remove old versions of functions if you're changing the arguments

express/whitelisted-queries.js
alexmiddeleer: Totally fine

next/components/deck-search-form.js
alexmiddeleer: Sounds good. If we did logging it would be good to log here but we don’t 😬

postgres/migrations/next-migration.sql
alexmiddeleer: Huh, that’s mildly concerning but we can ignore for now.

alexmiddeleer: Though maybe we should write a comment here to test this migration before deploying everything else

postgres/migrations/next-migration.sql
alexmiddeleer: Correct, else I think postgraphile gets confused

postgres/migrations/next-migration.sql
lily-pytel: Do you do migrations with adminer or command line? I think I was getting the same error with the original function on adminer as well

alexmiddeleer:
I do migrations via the command line. I’ll make sure to test this locally
before I deploy.

PR #293, opened 2020-08-13
Sort by deck views

lily-pytel:
Look at them views and sort by them!

image

image

### DIFF
--- a/postgres/migrations/next-migration.sql
+++ b/postgres/migrations/next-migration.sql
@@ -1,51 +1,3 @@
-CREATE OR REPLACE VIEW mythgard.deck_preview as
-  SELECT deck.id as deck_id,
-         deck.name as deck_name,
-         deck.created as deck_created,
-         mythgard.deck_factions(deck.id) as factions,
-         mythgard.deck_essence_cost(deck.id)::int as essence_cost,
-         mythgard.deck_votes(deck.id)::int as votes,
-         deck.archetype as deck_archetype,
-         deck.type as deck_type,
-         deck.modified as deck_modified,
-         account.id as account_id,
-         account.username as username,
-         account.account_type as account_type,
-         account.profile_icon_id as profile_icon_id,
-         mythgard.deck_hotness(deck.id)::int as hotness,
-         deck_views.views::int as views,
-         path.name as path_name,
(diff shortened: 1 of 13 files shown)


postgres/migrations/next-migration.sql
```diff
@@ -86,9 +38,19 @@ create or replace function mythgard.search_decks_nosort(
     -- type filter
     AND (typeFilter is NULL or deck.type = typeFilter)
     -- path filter
-    AND (pathName is NULL or path.name = pathName)
+    AND (
+      pathName is NULL or
+      pathName = '' or
+      (pathName = 'No path selected' and deck.path_id is null) or
```
alexmiddeleer, line 44: Not sure I understand this. I don't think it's possible for pathName to be a non-empty value and also have path_id be null. Or maybe it just doesn't occur in practice and this is defensive? either way it suggests we need an explanatory comment if this is correct. ditto for power.

lily-pytel: It's that special case of "no path selected" or "no power selected" in the dropdown. Somehow we have this in the prod database - not sure how the path and power tables were populated. Anyway, if that option is selected, we can decks that have no path\power. Does that make sense?

alexmiddeleer: Yep, but I don't believe path_id can be null. The no path selected path has an id.

lilypytel-toast: It's true, after this PR it can only have 3 values:

* empty string, which means don't filter by path at all (so show all decks)
* path name, which means filter to decks that have this path
* "no path selected", which means filter to decks that have no path

However, I would leave the null in case I messed up somewhere and it's still somehow sending null instead of `''`. It shouldn't though - I think I tested all cases! So I can get rid of it

lily-pytel: It's true, after this PR it can only have 3 values:

* empty string, which means don't filter by path at all (so show all decks)
* path name, which means filter to decks that have this path
* "no path selected", which means filter to decks that have no path

However, I would leave the null in case I messed up somewhere and it's still somehow sending null instead of `''`. It shouldn't though - I think I tested all cases! So I can get rid of it

PR #297, opened 2020-09-24
Add card set things

lily-pytel:
Started working on the simplest way to add the set filter to the cards and deck builder pages. Looks like all the images are under core, which is cool - less headache.

### DIFF
--- a/next/lib/card-queries.js
+++ b/next/lib/card-queries.js
@@ -14,6 +14,7 @@ export const getCardsQuery = () => {
       $manaGTE: Int
       $strengthGTE: Int
       $defenseGTE: Int
+      $cardset: String
     ) {
       cards(
         filter: {
@@ -23,6 +24,7 @@ export const getCardsQuery = () => {
                 { name: { includesInsensitive: $searchText } }
                 { rules: { includesInsensitive: $searchText } }
                 { subtype: { includesInsensitive: $searchText } }
+                { cardset: { includesInsensitive: $cardset } }
               ]
             }
             {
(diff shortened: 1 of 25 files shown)


alexmiddeleer:
Less headache for now at least 😸

next/lib/card-queries.js
```diff
@@ -120,7 +127,7 @@ export const executeCardQuery = (
       defenseGTE: defenseGTE || null,
       supertypes:
         supertypes && supertypes.length
-          ? supertypes.map(s => s.toUpperCase())
+          ? supertypes.map((s) => s.toUpperCase())
```
alexmiddeleer, line 130: Don’t need the parens

next/lib/card-queries.js
```diff
@@ -69,14 +72,14 @@ export const getCardsQuery = () => {
...
-export const intEnumsToGQLVars = manaCostEnums => {
+export const intEnumsToGQLVars = (manaCostEnums) => {
   if (!(manaCostEnums && manaCostEnums.length)) {
     return [null, null];
   }
 
   const discreteManaCosts = [];
   let manaGTE;
-  manaCostEnums.map(num => {
+  manaCostEnums.map((num) => {
```
alexmiddeleer, line 82: No parens needed. Your prettier version change?

alexmiddeleer (approved):
Wow you’re fast!

next/lib/card-queries.js
lily-pytel: New computer - I fixed it but forgot these files so thanks for the catch!

lily-pytel:
I love the vote of confidence but gonna add tests and stuff as well 😆

lily-pytel:
We might need to change the import card script since I just modified it blind but gonna merge anyway

PR #299, opened 2020-09-27
Hide spawns

alexmiddeleer:
Just the database pieces in this PR. Next PR will contain the graphql updates.

### DIFF
--- a/express/whitelisted-queries.js
+++ b/express/whitelisted-queries.js
@@ -275,12 +275,12 @@ mutation UpdateDeckAndRemoveCards(
           quantity
           card {
             name
+            cardset
             id
             mana
             gem
             supertype
             rarity
-            cardset
             cardFactions {
               nodes {
                 faction {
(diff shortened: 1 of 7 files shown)


alexmiddeleer:
turned out not to be so bad, so ignore what I said before - this PR has everything.

lily-pytel:
Interesting, so is there any way to see that spwan card?

alexmiddeleer:
@lily-pytel Only by clicking on a card that spawns something e.g. https://www.mythgardhub.com/card?id=329

lily-pytel:
Ah so does that card in the example database - is it attach to one of the other cards?

alexmiddeleer:
@lily-pytel nope not right now. I just wanted to create it so that I could verify it does not appear in searches

alexmiddeleer:
We already have a card that spawns other cards, but those spawns are not spawnOnly = true. Which could perhaps happen if a card spawns a true clone, such as https://scryfall.com/card/znr/203/scute-swarm (not sure if can happen in Mythgard but maybe someday)

lily-pytel (approved):
Looks great!

PR #300, opened 2020-10-13
Filtering bug fixes

lily-pytel:
- rarity was not being pulled into the filtered cards query
- card set was being treated as an OR filter instead of AND


alexmiddeleer (approved):
Cypress test - if feasible - would be ideal  but not required

lily-pytel:
I know, my cypress is failing on me! It started throwing an unfamiliar error 😢 😢 😢 . I'm gonna merge this so it's fixed and work on making cypress work again.

PR #306, opened 2020-11-09
Upgrade next

alexmiddeleer:


lily-pytel (approved):
Other than cypress going apeshit, looks good

lily-pytel:
Different issue here. For me none of the pages work (and neither do the cypress tests) with this error
image

@babel/core is there but I guess not being found properly

alexmiddeleer:
Hmm, that one is odd. Maybe docker is not setting up the image properly?
Did you force rebuild the images? Time to clean up our docker files, they
seem to suck.

lily-pytel:
I did and then I moved babel from devDeps to regular deps and the pages worked. Still getting hung on authentication but at least the pages worked. Also I destroy all docker with `docker system prune -a` every time so I don't think it's old docker files hanging around

lily-pytel:
When I take out line `passport.authenticate('google', { failureRedirect: '/', session: false }),` then all pages hang... hmmmmm. Perhaps the next router is the culprit?

lily-pytel:
In any case looks like this branch can be merged to tabs-for-mod-page. The upgrade seems successful and the builds are going to pass

lily-pytel:
Finally works for me though I see it failed the tests before... gonna check it out. In the meantime there's a little cosmetic issue where the selected tab is always the first one even though the content of the tab changes appropriately. Example:

image

The content belongs to the "Events" tab but the first tab still looks like it's selected

lily-pytel (approved):
Other than the small tab thing, everything looks great. Went through all the tabs and tried all the interactions (I hope) and they all seem to be working still.

lily-pytel:
Ah the failing build looks like the usual timeout. One of the Travises passed and the other failed so probably just flaky. Current builds are both passing.

PR #310, opened 2020-11-21
Spoilers

alexmiddeleer:
A very simple spoilers page. If you feel like manually testing this, try going to the mod page and adding either "Academy Analyst" or "A Mind of Its Own" on the spoilers tab. I will have another PR with a simple e2e test and modifications to init.sql so new apps have a spoiler or two.


lily-pytel (approved):
Just needed to bring all the hooks to the top (otherwise it throws an error) but everything worked great!

alexmiddeleer:
@lily-pytel thank you!

PR #317, opened 2020-12-02
Should be using npm ci for reliable images and ci

alexmiddeleer:


alexmiddeleer:
Interesting that postgraphile was completely wrong version in the lockfile. I really get annoyed with npm sometimes.

lily-pytel (approved):

PR #319, opened 2020-12-06
Replace failing tests

lily-pytel:
Can't believe this but it seems like even though `not.be.visible` works locally, on Travis the element actually doesn't exist??? So it looks like `not.exist` does work


lily-pytel:
Son of a 🔫 🔫 🔫 🔫 🔫 🔫 🔫

lily-pytel:
I am baffled as to why `not.be.visible` _used to work_ and not it doesn't. And why it works in development. Did Travis update? Do we run in dev mode locally? Is this just a cruel universe?

alexmiddeleer:
Wow can't believe the tests passed!

.travis.yml
```diff
@@ -33,7 +33,7 @@ before_install:
 install:
   # Install node modules (for jest) and start up the docker containers
   - cd next
-  - npm ci
+  - npm install
```
alexmiddeleer, line 36: why did we change this from ci? I think we should keep using ci....

e2e/cypress/integration/deck-builder-page.js
```diff
@@ -27,9 +27,9 @@ describe('Deck builder page', () => {
   it('should have a happy path', function() {
     cy.get('[data-cy="header"]').should('be.visible');
     cy.get('[data-cy="goToImportMode"]').should('be.visible');
-    cy.get('[data-cy="importDeckButton"]').should('not.be.visible');
-    cy.get('[data-cy="deckBuilderActions"]').should('not.be.visible');
-    cy.get('[data-cy="deckManaCurve"]').should('not.be.visible');
+    cy.get('[data-cy="importDeckButton"]').should('not.exist');
```
alexmiddeleer, line 30: Hmm, strange. Maybe cypress got upgraded and we didn't notice. Seems like a dumb breaking change though.

express/Dockerfile
```diff
@@ -4,7 +4,7 @@ WORKDIR /app
 
 COPY . /app
 
-RUN npm ci --production
+RUN npm i --production
```
alexmiddeleer, line 7: ditto I think we should keep using ci unless tests don't pass. If they don't pass I'm guessing we should try to figure out which version we need and keep using npm ci so it's less brittle.

alexmiddeleer (approved):
Hopefully the lib upgrades don't break production again but I'll roll back if they do. It's easy. Thanks so much for figuring out the tests, I was super stumped and ready to give up.

e2e/cypress/integration/deck-builder-page.js
lily-pytel: Very dumb!!! Can't believe this made the tests green

.travis.yml
lily-pytel: It looked like when ci was used, the tests failed in a different way but let me try bringing it back

lily-pytel:
Wow bringing back `npm ci` was _not_ well received by travis

lily-pytel:
What do you think about merging this? It's at least green and we can treat it as a starting point to bringing back `npm ci`.

alexmiddeleer:
@lily-pytel sounds good. Thanks again!

PR #325, opened 2021-01-23
Add tertiary faction

lily-pytel:
image


alexmiddeleer (approved):
Thanks, looks perfect

PR #330, opened 2021-03-15
Add the ability to filter by 0 cost, 0 attack or 0 defense

lily-pytel:
Small change to add 0 filters to the cards page. There are no 0 defense\health cards for now but... ya never know. 

image

The icon doesn't look great so I won't merge it until we have a better icon in aws though


alexmiddeleer (approved):
Thanks! Sorry for the slow review!

